PR #33, opened 2016-07-15
Fix the computation of fromForeignPtr, less segfaults, more smiles

ndmitchell:
If there is a length and an offset the end position is the offset plus length, so fix that.

As a side note, using explicit `Int#` in `UVectorAddr` is a pain, and probably doesn't buy anything over `{-# UNPACK #-} !Int` anyway.


vincenthz:
good catch !

Issue #35, opened 2016-07-21
FilePath API


vincenthz:
* [ ] extension manipulation
* [x] path dropping / append filename
* [ ] parent
* [ ] windows support
* [ ] OS support (`readdir`)

vincenthz:
linked with #27

NicolasDP:
Hi Vincent, if you don't mind I'll start working on this.

NicolasDP:
I'm thinking to keep what you already have done in the VFS (but in another module dedicated to Path).

PR #41, opened 2016-07-22
[WIP][#35] first draft implementation of Path/FilePath functions and API

NicolasDP:
design API of the Path/FilePath for ticket #35


vincenthz:
looking like a nice start for this brave new world
need to document the design choices, and limitation properly

Issue #36, opened 2016-07-21
Detecting operating system, cpu, related information


vincenthz:
Add a module and functions to:

* [x] detect operating system
* [x] detect cpu endianness
* [x] detect number cpus

NicolasDP:
👍

NicolasDP:
it'd be nice to have the compiler name and version

ndmitchell:
Is this correct for processor count? See what I do in Shake: https://github.com/ndmitchell/shake/blob/d45223407bba357e9e9498fd840b479581510697/src/General/Extra.hs#L56

Basically in the non threaded RTS number of CPU's == 1.

PR #42, opened 2016-07-23
add CPUs, Arch, compiler and Endianness detection

NicolasDP:
implements requirements for #36

### DIFF
--- a/Core/System/Info.hs
+++ b/Core/System/Info.hs
@@ -11,7 +11,8 @@
 module Core.System.Info
     (
       -- * Operation System info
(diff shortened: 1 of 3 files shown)


NicolasDP:
how sad this issue with ```cabal check``` and some architecture not being available on some versions...

vincenthz:
thanks ! a good start, probably a good idea to remove the C helper function that will pollute the library symbols exported and replace by the equivalent haskell

vincenthz:
future enhancement tracked as #45

Issue #39, opened 2016-07-22
Sequential, add intercalate and intersperse


NicolasDP:

JanGe:
Starting to work on it

PR #44, opened 2016-07-23
Add intersperse and intercalate to Sequential

JanGe:
Issue #39


vincenthz:
Thanks, and well done ! it turn out trickier than expected.

vincenthz:
I'll fix the issue with GHC 8.0 in a future commit

Issue #20, opened 2016-06-12
add zip, zipWith .. family of functions


vincenthz:

JanGe:
Starting to work on this.

PR #48, opened 2016-07-24
[#20] Add Zippable type class with zip and zipWith methods

JanGe:
Issue #20 

- [x] zip, zipWith, unzip
- [x] zip{2,7}, zipWith{2,7}, unzip{2,7}
- [x] Provide custom method definitions for `Array`, `UArray` and `String` with better performance
- [x] tests


Core/Array/Boxed.hs
```diff
@@ -108,6 +108,8 @@ instance C.IndexedCollection (Array ty) where
             | otherwise =
                 if predicate (unsafeIndex c i) then Just i else Nothing
 
+instance C.Zipable (Array ty) where
```
DavidM-D: I think it's spelt Zippable
https://en.wiktionary.org/wiki/zippable

JanGe:
Damn it. GHC HEAD won't let me make `String` an instance of `Zippable` as `zip` can't be used with it:

    Core/String/UTF8.hs:105:10: error:
    • Couldn't match type ‘Char’ with ‘(C.Element a, C.Element b)’
      Inaccessible code in
        the type signature for:
          C.zip :: (C.Sequential a, C.Sequential b,
                    C.Element String ~ (C.Element a, C.Element b)) =>
                   a -> b -> String
    • In the instance declaration for ‘C.Zippable String’

JanGe:
For tuples sizes >= 3 I didn't implement a type specific `zipWith` so far. Are they needed?

vincenthz:
wow @JanGe. didn't have time to read it all yet, but this seems great ! Will try to process over the weekend

JanGe:
Well, actually the `BoxedZippable` type class doesn't work yet for tuples sizes >= 3.

    $ zip3 (fromList "abc" :: String) (fromList "abc" :: String) (fromList "abc" :: String) :: Array (Char, Char, Char)

    <interactive>:5:1:
        Couldn't match type ‘(Char, Char, Char)’
                       with ‘(Element String, Element String)’
        In the expression:
            zip3
              (fromList "abc" :: String)
              (fromList "abc" :: String)
              (fromList "abc" :: String) ::
              Array (Char, Char, Char)
        In an equation for ‘it’:
            it
              = zip3
                  (fromList "abc" :: String)
                  (fromList "abc" :: String)

(And GHC HEAD catches that at compile time!)

Still thinking about how I could best solve this...

vincenthz:
Nice work. we can improve on the building side of things later on I think

PR #52, opened 2016-07-24
Fix missing base case in cons/snoc for UTF8 strings

JanGe:
Evaluating `cons` and `snoc` in `Core.String.UTF8` resulted in an exception when called with `mempty` while using `(fromList "" :: String)` worked fine:

    Core Core.Collection.Sequential> cons 'a' (mempty :: String)
    "*** Exception: empty de-referenced

    Core Core.Collection.Sequential> snoc (mempty :: String) 'a'
    "*** Exception: empty de-referenced

    Core Core.Collection.Sequential> cons 'a' (fromList "" :: String)
    "a"

    Core Core.Collection.Sequential> snoc (fromList "" :: String) 'a'
    "a"

Both failing functions called `copyAtRO` in `Core.Array.Unboxed` with a byte array pointer that can not be dereferenced.

QuickCheck tests didn't catch this as they currently only generate data using `fromList`.

`cons` and `snoc` now have a base case handling this. I also propose to rename `copyAtRO` to `unsafeCopyAtR0` as it does no bounds checking.

### DIFF
--- a/Core/Array/Boxed.hs
+++ b/Core/Array/Boxed.hs
@@ -469,20 +469,24 @@ mapIndex f a = create (length a) (\i -> f i $ unsafeIndex a i)
 -}
 
 cons ::  ty -> Array ty -> Array ty
-cons e vec = runST $ do
-    mv <- new (len + 1)
-    unsafeWrite mv 0 e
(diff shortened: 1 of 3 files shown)


vincenthz:
Good catch !

PR #53, opened 2016-07-24
implement the modified UTF8 code functions to try to by-pass the list of Char

### LINKED ISSUE #37: modified UTF-8 to string/bytearray
Add efficient way to convert GHC modified UTF-8 encoding to:

* [x] string
* [ ] bytearray

NicolasDP:
fix #37

### DIFF
--- a/Core/String/ModifiedUTF8.hs
+++ b/Core/String/ModifiedUTF8.hs
@@ -19,21 +19,19 @@ module Core.String.ModifiedUTF8
     ( fromModified
     ) where
 
-import           GHC.ST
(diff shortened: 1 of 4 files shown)


NicolasDP:
@vincenthz thx for the ByteArray builder. very useful and helped a lot to keep things simple

vincenthz:
it's sorely lacking some tests: having couple of hardcoded strings (one containing various unicode blocks, and null bytes) in the tests would be great.

vincenthz:
Thanks !

PR #56, opened 2016-07-26
better test on modified UTF8

NicolasDP:
test #37


vincenthz:
Looking good, please merge

Issue #59, opened 2016-08-05
arch should be an enum


ndmitchell:
What are typical values for arch? That's the obvious question when looking at the function.

NicolasDP:
This is a fair question. I guess it might be possible to dig into `base` to find the answer and document it.

ndmitchell:
For info, on Windows, i get `i386`.

NicolasDP:
I looked into base, I believe the potential results are: i386, x86_64, powerpc, sparc, arm...

https://github.com/ghc/ghc/blob/ffe4660510a7ba4adce846f316db455ccd91142a/mk/config.mk.in#L202-L203

ndmitchell:
I wonder if this should be an enum, much like OS?

vincenthz:
I would prefer as an enum too (eventually)

PR #67, opened 2016-08-05
update documentation of `arch`

NicolasDP:
address issue #59 

This is based on the information gathered from `ghc` and `base`.

In `GHC`, we can find the supported Arch:

* [in the config: TargetArch_CPP](https://github.com/ghc/ghc/blob/ffe4660510a7ba4adce846f316db455ccd91142a/mk/config.mk.in#L202-L203) for the build system;
* Then `HOST_ARCH` is defined with the value of `TargetArch_CPP` in [ghcplatform.h](https://github.com/ghc/ghc/blob/865602e0beb8e30ea1e1edf7db90f24088badb9e/includes/ghc.mk#L125-L139);
* which is then use for [System.Info.arch](https://github.com/ghc/ghc/blob/5d98b8bf249fab9bb0be6c5d4e8ddd4578994abb/libraries/base/System/Info.hs#L43)

### DIFF
--- a/Core/System/Info.hs
+++ b/Core/System/Info.hs
@@ -65,10 +65,14 @@ instance IsString OS where
 os :: OS
 os = fromString System.Info.os
 
--- | get the architecture info
+-- | get the machine architecture on which the program is running
 --
--- get the `arch` from base package but convert
--- it into a strict String
+-- This function uses base implementation:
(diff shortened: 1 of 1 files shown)


vincenthz:
Thanks

PR #71, opened 2016-08-06
Make `Arch` an enum

NicolasDP:
Also removed `Unknown` from `OS` type so it can be used for the `Arch` type.

address #59

### DIFF
--- a/Core/System/Info.hs
+++ b/Core/System/Info.hs
@@ -44,13 +44,11 @@ data OS
     | Linux
     | Android
     | BSD
-    | Unknown
     | Other String
   deriving (Show, Eq)
 
 instance IsString OS where
     fromString str = case str of
-        []        -> Unknown
         "darwin"  -> OSX
         "mingw32" -> Windows
         "linux"   -> Linux
--- a/Core/System/Info.hs
+++ b/Core/System/Info.hs
@@ -44,26 +44,24 @@ data OS
     | Linux
     | Android
     | BSD
-    | Other String
-  deriving (Show, Eq)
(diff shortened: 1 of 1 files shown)


Core/System/Info.hs
```diff
@@ -65,16 +65,46 @@ instance IsString OS where
...
+data Arch
+    = I386
+    | X86_64
+    | PowerPC
+    | PowerPC64
+    | PowerPC64LE
+    | Sparc
+    | Sparc64
+    | ARM
+    | AArch64
```
vincenthz: I'm not sure about the specific here.

I would call `AArch64` : `ARM64`, and get rid of PowerPC64LE (as this is fundamentally not an architecture). it should be the endian info that should be available someplace close that give this information

ndmitchell:
Why not OtherOS and OtherArch? That way they are consistent. Unknown and Other are just too generic names for very rarely used functionality. The other option would be to make the functions return Either String Arch, and elim Other/Unknown entirely? I think I'd prefer that... Enums plus one non enum just look a bit weird to me.

Core/System/Info.hs
NicolasDP: make sense. I'll remove that and make it clearer.

NicolasDP:
@ndmitchell : Make sense. The `IsString` is not really useful as well, I will remove it.

ndmitchell:
Looks good to me. Only remark is you could also derive Bounded. Plus having Data/Typeable on everything usually makes sense.

NicolasDP:
Good idea.

PR #73, opened 2016-08-07
add support for some different encoding formats

NicolasDP:
Now we can support some of the most common encoding:

1. ISO_8859_1
2. ASCII
3. UTF-16
4. UTF-32 (not tested though...)


vincenthz:
amazing. extremely solid effort ! thanks !

PR #77, opened 2016-08-14
Improve some docs in Foundation.VFS.Path

mitchellwrosen:


NicolasDP:
Thanks for the documentation improvements.

foundation.cabal
```diff
@@ -137,6 +137,13 @@ Test-Suite test-foundation
   ghc-options:       -Wall -fno-warn-orphans -fno-warn-missing-signatures
   default-language:  Haskell2010
 
+Test-Suite doctest
+  type:              exitcode-stdio-1.0
+  hs-source-dirs:    tests
+  Main-is:           DocTest.hs
+  Build-Depends:     base >= 3 && < 5
+                   , doctest >= 0.11
```
NicolasDP: in `lts-5.18`:

>  doctest: needed (>=0.11), 0.10.1 found

Do we specifically need, at least, this version of doctest? Another option would be to disable the `doctest` tests for lts-5 in Travis.

mitchellwrosen:
Oh, you do specifically need this version to automatically search directories, e.g. just pass "Foundation" to `doctest`. However, I didn't end up doing this because of `Foundation.String.UTF8LL`, which is an unused module, but confuses doctest nonetheless.

The version could be relaxed, or lts-5 could be given its own stack.yaml file.

vincenthz:
I've removed lts-5 tests in lights on the failure. it's not like it's telling us much since we don't have that much dependencies.

Thanks for the doctest patch, this is more than welcome !

Issue #51, opened 2016-07-24
GHCI issue (does not work due to unboxed tuple and bytecode interpreter)


NicolasDP:
this issue has a work around:

```
stack ghci --ghci-options '-fobject-code'
```

NicolasDP:
@JanGe we found the work arount to help with GHCI and Foundation.

The issue in GHCI is tracked [there](https://ghc.haskell.org/trac/ghc/ticket/1257)

vincenthz:
I wonder if we should put this in the README or maybe some wiki page. That will prevent it being pinned forever in the issue tracker ?

JanGe:
We could add the option to a project-wide `.ghci` file (see PR #87). This has [some drawbacks](https://downloads.haskell.org/~ghc/7.10.3/docs/html/users_guide/ghci-obj.html) for modules that would work interpreted though:

> There are disadvantages to compiling to object-code: you can't set breakpoints in object-code modules, for example. Only the exports of an object-code module will be visible in GHCi, rather than all top-level bindings as in interpreted modules.

vincenthz:
does it works with `stack ghci` as well ? if so, I'm happy with that.

JanGe:
Yes, tested it with `ghci` and `stack ghci`.

PR #87, opened 2016-08-17
Set `-fobject-code` option for GHCi project-wide

JanGe:
Workaround for #51


vincenthz:
I think that's better than putting a documentation workaround, thanks !

PR #88, opened 2016-08-19
export Arch

NicolasDP:
This will fix the issue that `Foundation.System.Info.arch` was not usable since the type `Arch` was not exported...


NicolasDP:
merging since this is only exporting what should have been exporting in 1st place.

PR #90, opened 2016-08-23
Implement parser for Sequential types

### LINKED ISSUE #89: Add parser combinators
I'm currently using attoparsec, but a foundation String based one would be a lot easier. Either wrappers over the attoparsec class, or a port to a new string type, or perhaps even a fresh parser combinator library in the same style.

NicolasDP:
Port the [memory](https:://github.com/vincenthz/hs-memory)'s Parser into `Foundation`.
This implementation keeps the polymorphic interfaces `memory` provides, but using any `Sequential` types.

fix #89


vincenthz:
Thanks. it's a useful starting point

PR #96, opened 2016-09-02
Fixed #94 (Add Bifunctor to Foundation)

### LINKED ISSUE #94: Add Bifunctor
Now Bifunctor is available by default in latest versions of base, it's probably a good idea to expose it; that also resolve `first` and `second`
https://hackage.haskell.org/package/base-4.9.0.0/docs/Data-Bifunctor.html

* [ ] Add compatibility for base < 4.8 up to whichever version we support (see foundation's `IsList` to see how it's done)
* [ ] expose `first` and `second` in `Foundation`

adinapoli:
I wouldn't be surprised if this PR needs amending both in terms of coding convention and package hierarchy conventions 😉 

I backported `Data.Bifunctor` from base using CPP for GHC older than 4.8.0.

### DIFF
--- a/Foundation.hs
+++ b/Foundation.hs
@@ -46,6 +46,7 @@ module Foundation
     , Prelude.Bounded (..)
     , Prelude.Enum (..)
     , Prelude.Functor (..)
+    , Foundation.Class.Bifunctor.Bifunctor (..)
     , Control.Applicative.Applicative (..)
     , Prelude.Monad (..)
     , (Control.Monad.=<<)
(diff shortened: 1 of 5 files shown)


vincenthz:
I'm not sure about the location of this (we don't have a precedent except from IsList which is in Internal for giving a compat layer). I think it should be just be available in the Prelude (Foundation) otherwise.

Otherwise looking spot on what I expected

adinapoli:
Ok, changed pushed, let's wait for travis to validate 😀

NicolasDP:
looking good. just the test that needs to be fixed.

adinapoli:
Sorry guys, had no clue that you were using `second` from `Control.Arrow` in the tests and that the import would clash! Should be all green now.

NicolasDP:
Good stuff

PR #97, opened 2016-09-02
expose copy operation for Boxed Array and Strings

### LINKED ISSUE #83: array / string : copy operation
Specially relevant for things that contains slices, it's important to be able to remove reference to another memory bits so that they become independent.

vincenthz:
fix #83


NicolasDP:
nice. good stuff

PR #99, opened 2016-09-03
[WIP] Fixed #98, add Coveralls integration

### LINKED ISSUE #98: Integrate Coveralls
It would definitely be nice to have a badge displaying the current code coverage of the tests as generated by `hpc`.

In the past [I have used](https://github.com/adinapoli/threads-supervisor/blob/master/.travis.yml#L29) this [project](https://github.com/rubik/stack-hpc-coveralls) with success, and should be very easy to integrate.

I will whip up something.

adinapoli:
I copied verbatim the options from my other repo, so that should do the trick, but I think we
will need to explicitly enable Foundation in [Coveralls](https://coveralls.io).

I think this requires one of the owner of the foundation repo (probably Vincent) to sign up and add the project, as I (as a member) can only enable _my own fork_, not the real deal:

screen shot 2016-09-03 at 09 39 46

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -133,7 +133,7 @@ install:
   set -ex
   case "$BUILD" in
     stack)
-      stack --no-terminal --install-ghc $ARGS test --bench --only-dependencies
+      stack --no-terminal --install-ghc $ARGS test --coverage --bench --only-dependencies
       ;;
     cabal)
       cabal --version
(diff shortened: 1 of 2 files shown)


NicolasDP:
I have added it to coverall, is it still WIP ? https://coveralls.io/github/haskell-foundation/foundation

adinapoli:
@NicolasDP Usually once Coveralls is fully set we should get it to post on each new open PR and get a new check, as below:

screen shot 2016-09-03 at 16 52 59

screen shot 2016-09-03 at 16 53 13

I suspect though, that since we added the permissions at a later stage we won't see those things now. Probably it's just easier to merge this and add small incremental tweaks until we get it to work 😉

PR #108, opened 2016-09-03
Use as Coverage badge the one endorsed by Coveralls.

adinapoli:
Coverall seems to think we do not have a badge on the homepage:

screen shot 2016-09-03 at 17 02 34

I think that was just either the fact it's being slow to catch up, or that the "alt" message in the markdown was "Coverage" and not "Coverage Status". Either way, this PR should fix it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ Foundation
 ==========
 
 [![Build Status](https://travis-ci.org/haskell-foundation/foundation.png?branch=master)](https://travis-ci.org/haskell-foundation/foundation)
-[![Coverage](https://coveralls.io/repos/github/haskell-foundation/foundation/badge.svg?branch=master)](https://coveralls.io/github/haskell-foundation/foundation?branch=master)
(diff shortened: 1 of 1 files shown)


adinapoli:
On a side note: the coverage report seems to be written on Travis, although there is an error returned (by stack):

screen shot 2016-09-03 at 17 53 03

NicolasDP:
There are also issues when `GHCVER` is `head` and we can also see issues with other versions:

https://travis-ci.org/haskell-foundation/foundation/jobs/157321835

```
0.88s$ ./shc foundation test-foundation
ghc-pkg: cannot find package foundation
shc: readCreateProcess: stack "exec" "--" "ghc-pkg" "field" "foundation" "key" "--simple-output" (exit 1): failed
```

adinapoli:
@NicolasDP I'm using a fairly outdated version of `shc` as it was the one I had success with, but I had the burning suspect it's degree of effectiveness will be dependant by the byproduct (stack, ghcVer). In other words, I think that using a particular permutation of {shc,stack,ghc} is not always guaranteed to work, just due to the brittle nature of these tools, which probably relies on some particular stacks's command line options to be present.

NicolasDP:
it makes sense

PR #110, opened 2016-09-03
ASCII String type

NicolasDP:
This is only a simple wrapper on top of UArray Word8.

Very useful type specialisation to write parser of object that are limited to ASCII characters...

### DIFF
--- a/Foundation/String/ASCII.hs
+++ b/Foundation/String/ASCII.hs
@@ -1,0 +1,305 @@
+-- |
+-- Module      : Foundation.String.ASCII
+-- License     : BSD-style
+-- Maintainer  : Haskell Foundation
+-- Stability   : experimental
+-- Portability : portable
+--
+-- A AsciiString type backed by a `ASCII` encoded byte array and all the necessary
+-- functions to manipulate the string.
+--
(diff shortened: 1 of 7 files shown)


Foundation/String/ASCII.hs
```diff
@@ -0,0 +1,305 @@
...
+
+newtype MutableAsciiString st = MutableAsciiString (MutableByteArray st)
+    deriving (Typeable)
+
+instance Show AsciiString where
+    show = show . sToList
+instance IsString AsciiString where
+    fromString = sFromList
+instance IsList AsciiString where
+    type Item AsciiString = Char
```
vincenthz: `Char` should be `CChar`

Foundation/String/ASCII.hs
```diff
@@ -0,0 +1,305 @@
...
+    fromList = sFromList
+    toList = sToList
+
+type instance C.Element AsciiString = Char
```
vincenthz: same here

vincenthz:
Thanks. we'll have to see if we can make it useful (e.g. parsing ASCII like data more conveniently). I'm a bit dubious it's going to work

PR #116, opened 2016-09-10
Add support for Non empty as a property

### LINKED ISSUE #113: FIP: Add the NonEmpty property type wrapper
`NonEmpty`is currently used by `semigroups` recently introduced in `base`. it's defined as a normal cons without a empty constructor:

```haskell
data NonEmpty a = a :| [a]
```

Proposing a NonEmpty property to replace it:

```haskell
newtype NonEmpty a = NonEmpty a

nonEmpty :: Collection a => a -> Maybe (NonEmpty a)
nonEmpty a | null a = Nothing
           | otherwise = Just a
-- Maybe could be a Partial, or we could add nonEmpty_ that throws ...
```

Advantages:

* Works for any collection: `NonEmpty (UArray Word8)`, `NonEmpty String`
* use a smart constructor to make sure only non empty collection are created

As a disadvantage:

* Doesn't deconstruct the first element by construction like the current one, so appear less inherently safe. e.g. `data NonEmpty a = NonEmpty a (Collection a)`

This open the door for many functions to become safer and non partial (but *not* by construction):

```haskell
minimum :: Collection a => NonEmpty a -> Element a
maximum :: Collection a => NonEmpty a -> Element a
head :: Collection a => NonEmpty a -> Element a
last :: Collection a => NonEmpty a -> Element a
uncons :: Collection a => NonEmpty a -> (Element a, a)
foldl1 ...
foldl1' ...
foldr1 ...
```

vincenthz:
Also:

* remove Monoid from Collection
* define minimum,maximum
* define head,last,init,tail

fix #113

### DIFF
--- a/Foundation/Array/Bitmap.hs
+++ b/Foundation/Array/Bitmap.hs
@@ -79,6 +79,8 @@ instance C.Foldable Bitmap where
 instance C.Collection Bitmap where
     null = null
     length = length
+    minimum = Data.List.minimum . toList . C.getNonEmpty -- TODO can shortcircuit all this massively
+    maximum = Data.List.maximum . toList . C.getNonEmpty -- TODO DITTO
 instance C.Sequential Bitmap where
     take = take
     drop = drop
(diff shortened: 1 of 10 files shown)


NicolasDP:
great stuff.

PR #133, opened 2016-09-27
Add hashing capability

vincenthz:
* [x] FNV hash family
* [x] Siphash
* [x] Define a class of hashable data type
* [x] Add instance for all basic types

### DIFF
--- a/Foundation/Hashing.hs
+++ b/Foundation/Hashing.hs
@@ -1,0 +1,21 @@
+-- |
+-- Module      : Foundation.Hashing
+-- License     : BSD-style
+-- Maintainer  : Foundation
+--
+module Foundation.Hashing
+    ( Hashable(..)
+    , Hasher
+    -- * Specific methods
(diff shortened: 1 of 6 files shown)


ndmitchell:
@vincenthz you are an expert at hash functions, so could I ask for well named "obvious" things for us unknownedgeable users? Specifically something named hash or hashCrypto or similar which is a good balance of safety/security for most users. I doubt most users are able to correctly pick FNV vs SIP.

vincenthz:
@ndmitchell completely agreed. While FNV and siphash will be explicit choice that one can make, the plan is to have 2 defaults choice (one being the DoS resistant, and the one being full speed hashing) which hide the implementation choice of FNV vs siphash (or any later hash functions)

vincenthz:
incomplete in term of the instances (Int, Integer, IntN missing mainly) but merging to get to a newer baseline

PR #163, opened 2016-12-06
add pipe & conduit types

vincenthz:
Experiment bringing necessary conduits types and functions to foundation. mostly Types copied from `conduit` package.


vincenthz:
adding @snoyberg

snoyberg:
We have plans for the future to s/`ConduitM`/`Conduit`, so you may want to consider doing that now. Also, arguably the constructor names should be changed for `HaveOutput`=>`Yield` and `NeedInput`=>`Await`.

vincenthz:
great ! I was really thinking that already when tweaking the code around. I think that make sense to match the functions that construct them

vincenthz:
BTW, The overall goals also is to provide all the machinery by default in a conduit-only-flavor.
so all file reading/writing will be done exclusively through conduit API.

snoyberg:
The build is failing on older GHCs because older base does not include `Data.Functor.Identity`. I _think_ it makes the most sense to export `Identity` from `Foundation.Base.Monad`, either using the version from `base` or a locally defined one. Does that make sense to you @vincenthz? Alternative is I can just drop the `runConduitPure` function.

Semi-related question is whether I should add a full `ResourceT` implementation. Lots of real-world conduit code uses it, so I'd be in favor of it overall unless there's some objection.

vincenthz:
yes, let's add a proper foundation `Identity` with the right compat in one place. I don't think it "cost" much to do that.

Related to resourcet, I'm not very familiar of the design space to make an inform choice right now; is that roughly akin to a resource finalizer that is synchronously run by the monadic code ? In any case, I'm happy to trust your judgment, if you think adding `ResourceT` is the way to go, then let's do that.

snoyberg:
Let's do the `ResourceT` stuff in a separate PR and/or issue. I think it makes sense, but it will have some other impacts, such as needing to add a `MonadMask` or similar typeclass (`MonadCatch` is insufficient). Also, this is a good cutting point for conduit functionality, since it allows defining lots of streaming operations, but just requires allocating scarce resources outside the pipeline.

vincenthz:
Thanks @snoyberg !

PR #169, opened 2016-12-15
add instance of `Storable` for `Ptr`

NicolasDP:
for now we simply re-use the `base` ones.

### DIFF
--- a/Foundation/Class/Storable.hs
+++ b/Foundation/Class/Storable.hs
@@ -21,6 +21,7 @@ import GHC.Types (Double, Float)
 
 import Foreign.Ptr (castPtr)
 import qualified Foreign.Ptr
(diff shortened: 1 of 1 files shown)


Foundation/Class/Storable.hs
```diff
@@ -124,3 +128,9 @@ instance StorableFixed Word32 where
 instance StorableFixed Word64 where
     size      = primSizeInBytes . toProxy
     alignment = primSizeInBytes . toProxy
+instance StorableFixed (Ptr a) where
+    size      = Size . Foreign.Storable.sizeOf    . toUndefined
```
vincenthz: probably easier to use directly here `(undefined :: Proxy a -> a)` instead of defining `toUndefined` globally to the module

PR #171, opened 2016-12-17
Older versions of `base` don't have `GHC.Char`

### LINKED ISSUE #170: inaccurate version bounds / compile failure
.cabal file specifies support for `base == 4.*` and `ghc-prim == *`; however, GHC 7.4.2 disagrees:

```
Configuring component lib from foundation-0.0.2
Preprocessing library foundation-0.0.2...

Foundation/String/UTF8.hs:62:18:
    Could not find module `GHC.Char'
    Perhaps you meant GHC.MVar (from base)
    Use -v to see a list of the files searched for.
```

NicolasDP:
fix #170 

In `String.Utf8` we uses `eqChar` from `GHC.Char` (only available from `base-4.9.0`)

This is the [existing code](https://github.com/haskell-foundation/foundation/blob/master/Foundation/String/UTF8.hs#L691-L695) making is of `GHC.Char.eqChar`:

```haskell
#if MIN_VERSION_base(4,9,0)
{-# RULES "break (== 'c')" [3] forall c . break (eqChar c) = breakElem c #-}
#else
{-# RULES "break (== 'c')" [3] forall c . break (== c) = breakElem c #-}
#endif
```

This fix does the same and includes `GHC.Char` **if** `base` is at least version `4.9.0`.

### DIFF
--- a/Foundation/String/UTF8.hs
+++ b/Foundation/String/UTF8.hs
@@ -59,7 +59,9 @@ import           GHC.Prim
 import           GHC.ST
 import           GHC.Types
 import           GHC.Word
+#if MIN_VERSION_base(4,9,0)
 import           GHC.Char
+#endif
 
  -- temporary
 import qualified Data.List


vincenthz:
looking good

PR #172, opened 2016-12-19
disable bench for old version of ghc/cabal

NicolasDP:
Simply don't install bench dependencies for the olf versions of cabal.

This fixes the travis tests.


vincenthz:
this doesn't really scale. why no add a RUNBENCH=0 just like RUNTEST=0

PR #174, opened 2016-12-28
Foundation's Socket and network

NicolasDP:
# Overview

Provide a simple yet more modern interface for network operaitons.

## Keep Socket type, family and protocol

This will provide a better user experience: a `Socket` will embed the `Family`, the `Type` and the `Protocol`. This can be used later to enable interfaces interfaces or behaviours on a `Socket depending on the `Family`, the `Type` or the `Protocol`.

```haskell
newtype Socket socket_family socket_type socket_protocol
```

## Integrated with the **Foundation**'s primitive types

`send` function accepts a `UArray ty`. We could extend it to take any `Foldable`  and then write GHC RULES for specific cases (like `UArray`).

# Missing

* [x] Windows internals
* [ ] documentation is poor
* [ ] automatic tests
* [ ] would be good to use `GHC.Conc.threadWaitReadSTM` and `GHC.Conc.threadWaitWriteSTM`
* [ ] streamable interfaces (for when the `socket_type` is `Stream`)

PR #187, opened 2017-01-18
Add Network's type: IPv4

NicolasDP:
This PR is part of the work of splitting #174 into smaller PRs.

## IPv4

Simply add the type, the tests and provide the basic interfaces.

## Convenient other changes

* re-export the class `Bits` (from `base`) from `Foundation.Bits`;
* add some compatibility function (`byteSwapXX`) as not implemented before `base` **4.7.0**;

### DIFF
--- a/Foundation/Bits.hs
+++ b/Foundation/Bits.hs
@@ -5,11 +5,16 @@ module Foundation.Bits
     , Bits(..)
     , alignRoundUp
     , alignRoundDown
+    , htonl, htons
+    , ntohl, ntohs
     ) where
 
 import Foundation.Internal.Base
+import Foundation.Internal.ByteSwap
 import Foundation.Numerical
+import Foundation.System.Info (Endianness(..), endianness)
 import Data.Bits
+import Data.Word (Word16, Word32)
 
 -- | Unsafe Shift Left Operator
 (.<<.) :: Bits a => a -> Int -> a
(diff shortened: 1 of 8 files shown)


Foundation/Internal/ByteSwap.hs
```diff
@@ -0,0 +1,66 @@
...
+
+-- | compatibility implementation (i.e. may be slow)
+--
+-- Swap the bytes position (inverting order) as would be done in GHC >= 7.8
+byteSwap32 :: Word32 -> Word32
+byteSwap32 w = w1 .|. w2 .|. w3 .|. w4
+  where
+    w1,w2,w3,w4 :: Word32
+    w1 = (w `unsafeShiftL` 24) .&. 0xFF000000
+    w2 = (w `unsafeShiftR` 24) .&. 0x000000FF
```
vincenthz: are you sure it's not 16 here ? :trollface:

Foundation/Bits.hs
```diff
@@ -39,3 +45,27 @@ alignRoundDown :: Int -- ^ Number to Align
                -> Int -- ^ Alignment (power of 2)
                -> Int
 alignRoundDown m alignment = m .&. complement (alignment-1)
+
+-- | perform conversion from Host to Network endianness
+htons :: Word16 -> Word16
+htons w = case endianness of
+    LittleEndian -> byteSwap16 w
+    BigEndian    -> w
```
vincenthz: I would really like to get away from host/network terminology (which only really apply to ipv4 in our case, as ipv6 do something different). I would much prefer to just have Little Endian and Big Endian memory abstraction

Foundation/Network/IPv4.hs
```diff
@@ -0,0 +1,76 @@
...
+  where
+    f = fromIntegral
+    w1, w2, w3, w4 :: Word32
+    w1 = w .>>. 24 .&. 0x000000FF
+    w2 = w .>>. 16 .&. 0x000000FF
+    w3 = w .>>.  8 .&. 0x000000FF
+    w4 = w         .&. 0x000000FF
+
+
+foreign import ccall unsafe "inet_addr"
```
vincenthz: That should done in haskell not a binding to C. this is not terribly difficult specially ipv4.

PR #189, opened 2017-01-22
Add ResourceT for conduit usage

snoyberg:
This pulled in a few more features

* MonadBracket, to allow for a generalized way to install cleanups
* ResourceT/MonadResource
* runConduitRes and bracketConduit
* The sourceFile and sinkFile functions, demonstrating usage


vincenthz:
It would probably make sense to move resourceT outside of conduit later one, but that's great otherwise ! Thanks

PR #199, opened 2017-02-26
Make Float and Double as the instance of Signed.

sighingnow:
The `abs` method is necessary for Float and Double type.


NicolasDP:
thanks for adding these.

PR #200, opened 2017-02-26
add Big Endian and Little Endian types

NicolasDP:
This concept is imported from vincenthz/hs-memory. The idea is to be able to add information about the byte ordering on the type. This will replace the `htonl` and co from `Foundation.Bits`.

### DIFF
--- a/Foundation/Primitive/Types.hs
+++ b/Foundation/Primitive/Types.hs
@@ -7,6 +7,7 @@
 --
 {-# LANGUAGE MagicHash #-}
 {-# LANGUAGE UnboxedTuples #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE CPP #-}
 module Foundation.Primitive.Types
     ( PrimType(..)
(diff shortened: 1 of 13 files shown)


Foundation/Primitive/Types.hs
```diff
@@ -219,6 +221,36 @@ instance PrimType Word16 where
     {-# INLINE primAddrRead #-}
     primAddrWrite addr (Offset (I# n)) (W16# w) = primitive $ \s1 -> (# writeWord16OffAddr# addr n w s1, () #)
     {-# INLINE primAddrWrite #-}
+instance PrimType (BE Word16) where
```
vincenthz: I think that `instance PrimType a => PrimType (BE a) where` (and same for LE) would have worked instead of explicitely adding each instances manually. `BE` and `LE` are just tag to represent that the value held inside is a different endianess. technically you could very well have a `BE Natural`, `BE Word128`, `BE CUInt` etc..

PR #204, opened 2017-03-07
IPv6 and IPv4 parsers

NicolasDP:
Removes the IPv4's FFI for `inet_addr` it was used for string parsing.

IPv4 and IPv6 now have a better string parsing using Foundation's Parser.

### DIFF
--- a/Foundation/Network/IPv4.hs
+++ b/Foundation/Network/IPv4.hs
@@ -1,4 +1,3 @@
-{-# LANGUAGE ForeignFunctionInterface #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GeneralizedNewtypeDeriving #-}
 
@@ -6,12 +5,10 @@ module Foundation.Network.IPv4
     ( IPv4
     , fromString, toString
     , fromTuple, toTuple
+    , ipv4Parser
     ) where
 
-import Prelude (fromIntegral)
-
-import System.IO.Unsafe (unsafePerformIO)
-import Foreign.C.String (withCString, CString)
+import Prelude (fromIntegral,read)
 
 import Foundation.Class.Storable
 import Foundation.Hashing.Hashable
@@ -20,6 +17,8 @@ import Foundation.Internal.Proxy
 import Foundation.String (String)
 import Foundation.Primitive
 import Foundation.Bits
+import Foundation.Parser
+import Foundation.Collection
 
 -- | IPv4 data type
 newtype IPv4 = IPv4 Word32
@@ -39,8 +38,7 @@ toString :: IPv4 -> String
 toString = fromList . toLString
 
 fromLString :: [Char] -> IPv4
-fromLString str = unsafePerformIO $ withCString str $ \cstr ->
-    IPv4 . fromBE <$> c_inet_addr cstr
+fromLString = parseOnly ipv4Parser
 
 toLString :: IPv4 -> [Char]
 toLString ipv4 =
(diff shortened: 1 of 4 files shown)


Foundation/Network/IPv4.hs
```diff
@@ -72,6 +70,21 @@ toTuple (IPv4 w) =
     w3 = w .>>.  8 .&. 0x000000FF
     w4 = w         .&. 0x000000FF
 
-
-foreign import ccall unsafe "inet_addr"
-    c_inet_addr :: CString -> IO (BE Word32)
+-- | Parse a IPv4 address
+ipv4Parser :: (Sequential input, Element input ~ Char) => Parser input IPv4
+ipv4Parser = do
+    i1 <- takeAWord8
```
vincenthz: Using an applicative style in the future would be nice something like:

```
i2 <- element '.' <* takeAWord8
...
```

Foundation/Parser.hs
```diff
@@ -276,3 +308,76 @@ skipWhile p = Parser $ \buf err ok ->
 -- stream
 skipAll :: Sequential input => Parser input ()
 skipAll = Parser $ \buf err ok -> runParser flushAll buf err ok
+
+data Count = Never | Once | Twice | Other Int
```
vincenthz: `Word`

vincenthz: I'm really not convinced this is worth it to name special cases like Never, Once, Twice. Just replacing the whole `Count` by a `Word` seems more flexible and less polluting to the parsing namespace.

NicolasDP:
hm, again some issue with the travis build on osx which timeout. Merging anyway as one of the 2 job succeed.

Foundation/Parser.hs
```diff
@@ -276,3 +308,76 @@ skipWhile p = Parser $ \buf err ok ->
...
+-- If you know your parser must performs from 0 to 8 times:
+--
+-- ```
+-- repeat (Between Never (Other 8))
+-- ```
+--
+-- *This interface is still WIP* but went handy when writting the IPv4/IPv6
+-- parsers.
+--
+repeat :: Sequential input => Condition -> Parser input a -> Parser input [a]
```
vincenthz: This really looks like `replicateM` with a more flexible amount of repetition (variable). is it really worth it though.

for example in the parsing the ipv4 digits, where you expect `Bound Once (Other 3)`, you need to check after the fact that it's not > 255 (e.g. `127.0.0.258`), so you still need to check after parsing that the number fits. why not just `take1While isDigit` and check that it's valid just like you need to do anyway ?

Foundation/Network/IPv6.hs
```diff
@@ -155,3 +135,122 @@ toTuple (IPv6 hi low) =
...
+    when (len < 1) $ fail "invalid compressed IPv6 addressed"
+    return $ bs1 <> replicate len 0 <> bs2
+
+skipColon :: (Sequential input, Element input ~ Char)
+          => Parser input ()
+skipColon = element ':'
+skipDot :: (Sequential input, Element input ~ Char)
+        => Parser input ()
+skipDot = element '.'
+takeAWord8 :: (Sequential input, Element input ~ Char)
```
vincenthz: name is a bit misleading, maybe `takeDecimal8` ?

Foundation/Network/IPv6.hs
```diff
@@ -155,3 +135,122 @@ toTuple (IPv6 hi low) =
...
+           => Parser input Word16
+takeAWord8 = do
+    read <$> repeat (Between Once (toEnum 3)) (satisfy isDigit)
+takeAWord16 :: (Sequential input, Element input ~ Char)
```
vincenthz: same here, takeHexadecimal4 ?

Foundation/Network/IPv6.hs
```diff
@@ -155,3 +135,122 @@ toTuple (IPv6 hi low) =
...
+--
+-- ```
+--     ipv6ParserPreferred
+-- <|> ipv6ParserIPv4Embedded
+-- <|> ipv6ParserCompressed
+-- ```
+--
+ipv6Parser :: (Sequential input, Element input ~ Char)
+           => Parser input IPv6
+ipv6Parser =  ipv6ParserPreferred
```
vincenthz: I'm not sure it's a good idea to use `Alternative` here and instead not relying on (much slower) backtracking (restarting parsing at the beginning everytime). It's fine right now though.

PR #216, opened 2017-03-18
HostName and getHostNameInfo

NicolasDP:
add function to retrieve addresses from a given hostname

```haskell
addrs <- (addresses <$> getAddressInfo "github.com") `catch` \_ -> return [] :: IO [IPv4]
case addrs of
    [] -> error "no knonw address"
    i:is -> putStrLn $ "IPv4 " <> toString i
```

- [ ] missing test
- [x] move the c_get_h_errno in the Foundation's binding
- [x] use `safe` instead of `unsafe` for the gethostbyname2 FFI


Foundation/Network/HostName.hsc
```diff
@@ -0,0 +1,166 @@
...
+#include <winsock2.h>
+#else
+#include "netinet/in.h"
+#include "netdb.h"
+#endif
+
+-- | HostName
+--
+newtype HostName = HostName { toString :: String }
+  deriving (Show, Eq, Ord, Typeable, Hashable)
```
vincenthz: I think it make sense to not derive Show automatically, as this is a primitive type. hostname should be displayed as if it was a String itself

Foundation/Network/HostName.hsc
```diff
@@ -0,0 +1,166 @@
...
+-- re-entrant function `gethostbyname2`. This function is using a `MVar ()` to
+-- avoid a race condition and should be safe to use.
+--
+getHostNameInfo :: (Show address_type, Eq address_type, Address address_type)
+                => HostName
+                -> IO (HostNameInfo address_type)
+getHostNameInfo = getHostNameInfo_ Proxy
+
+globalMutex :: MVar ()
+globalMutex = unsafePerformIO (newMVar ())
```
vincenthz: probably missing a NOINLINE pragma

Foundation/Network/HostName.hsc
```diff
@@ -0,0 +1,166 @@
...
+        return $ HostNameInfo (HostName on) as addrs
+  where
+    loop f = do
+        ptr <- f
+        if ptr /= nullPtr
+            then return ptr
+            else do
+                err <- c_get_h_errno
+                case err of
+                    _ | err == (#const HOST_NOT_FOUND) -> throwIO $ HostNotFound h
```
vincenthz: errno are available from `Foundation.System.Bindings.Posix`

NicolasDP: this is not errno

vincenthz: yep, didn't really noticed the h. it really need a better bindings, and moving to System.Bindings too.

Foundation/Network/HostName.hsc
NicolasDP: thanks

Foundation/Network/HostName.hsc
NicolasDP: Good idea

Foundation/Network/HostName.hsc
```diff
@@ -0,0 +1,166 @@
...
+             -> IO [address_type]
+getAddresses p ty len ptr = do
+    x <- peek ptr
+    if x == any
+        then return []
+        else (:) x <$> getAddresses p ty len (ptr `plusPtr` 1)
+
+foreign import ccall unsafe "gethostbyname2"
+    c_gethostbyname2 :: CString -> CInt -> IO (Ptr Word8)
+foreign import ccall unsafe "get_h_errno"
```
NicolasDP: not really, it is not the same error code (`errno /= h_errno`). I will move the code in the Foundation's Sys binding.

Foundation/Network/Address.hsc
```diff
@@ -0,0 +1,53 @@
...
+
+#ifdef mingw32_HOST_OS
+#include <winsock2.h>
+#else
+#include "netinet/in.h"
+#endif
+
+-- | Network Address class
+--
+class StorableFixed addr => Address addr where
```
vincenthz: I don't think `Address` is useful

Foundation/Network/Address.hsc
```diff
@@ -0,0 +1,53 @@
...
+    -- | retrieve associated Network Family code for this type of address
+    familyCode :: proxy addr -> CInt
```
vincenthz: nothing should expose a family code directly

Foundation/System/Bindings/Network.hsc
```diff
@@ -0,0 +1,45 @@
...
+import Foreign.C.Types
+
+#ifdef mingw32_HOST_OS
+# include <winsock2.h>
+#else
+# include "netinet/in.h"
+# include "netdb.h"
+#endif
+
+herr_HostNotFound
```
vincenthz: this doesn't follow the sys<X><FCT> naming from other modules.

vincenthz: or `sys<X>_<DEFINE>`

NicolasDP: I see where you come from but this is not really related to `errno`. Would rather like to avoid confusion, especially for something restricted to hostname and that would certainly not have any other usage in other modules ?

cbits/foundation_network.c
```diff
@@ -0,0 +1,16 @@
+#include "foundation_system.h"
+
+#if defined(FOUNDATION_SYSTEM_WINDOWS)
+# include <winsock2.h>
+#else
+# include "netdb.h"
+#endif
+
+int get_h_errno(void)
```
vincenthz: if it's exposed it ought to be called `foundation_`.
also if WSAGetLastError, bindings to it directly (`sysWindowsSAGetLastError`) and preventing a needless export on windows sounds useful.

foundation.cabal
```diff
@@ -110,6 +112,7 @@ Library
                      Foundation.Collection.Foldable
                      Foundation.Collection.Mutable
                      Foundation.Collection.Zippable
+                     Foundation.Collection.Mapable
```
vincenthz: Mappable

NicolasDP:
The feature is working and should be safe to use for now. However, there will certainly be a change soon as I will explore other options with `getaddrinfo`.

PR #226, opened 2017-04-02
Tests

vincenthz:
* [ ] Arbitrary Integer
* [ ] Arbitrary Natural
* [ ] Arbitrary Bool
* [ ] Some other arbitrary
* [ ] printing errors through `show` in `forAll`
* [ ] general pretty printing of log
* [ ] some property definition (get inspiration from quickcheck)
* [ ] Add flag parsing from command line in defaultMain (e.g. number of tests per property by default)
* [ ] Use random seed instead of hardcoded seed
* [ ] Add ability to get user defined seed from user

Optional:

* [ ] add parallelism

### DIFF
--- a/Foundation/Check.hs
+++ b/Foundation/Check.hs
@@ -138,14 +138,20 @@ testName (Unit s _)     = s
 testName (Property s _) = s
 testName (Group s _)    = s
 
+type Context = (Word, [String])
+
 -- | Run tests
 defaultMain :: Test -> IO ()
 defaultMain test = do
-    runTest test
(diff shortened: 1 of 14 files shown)


NicolasDP:
@vincenthz it seems it is going to be a big chunk of work to review once the TODO is done.
What about we merge as is and we move your todo list in the Foundation `test` project ?

NicolasDP:
I have added your TODO list to the project Testing.
I think we can merge what you have and keep smaller PR for next time.

PR #236, opened 2017-04-06
String read

### LINKED ISSUE #231: Add readDouble
A function String to Double that parsed it would be great. Perhaps a few varieties with progressively more features - e.g. One on 0.0, one for 0.0e0, one with Nan/inf.

vincenthz:
* [x] Add readFloatingExact
* [x] Convert FloatingExact to Float / Double
* [x] Add readDouble

fix #231

### DIFF
--- a/Foundation/String/UTF8.hs
+++ b/Foundation/String/UTF8.hs
@@ -1248,6 +1248,9 @@ builderBuild sizeChunksI sb
         Vec.unsafeCopyAtRO mba (sizeAsOffset (end - sz)) x (Offset 0) sz
         fillFromEnd (end - sz) xs mba
 
+-- | Read an Integer from a String
+--
+-- Consume an optional minus sign and many digits until end of string.
 readInteger :: String -> Maybe Integer
 readInteger str
     | sz == 0  = Nothing
@@ -1262,6 +1265,9 @@ readInteger str
   where
     !sz = size str
 
+-- | Read a Natural from a String
+--
+-- Consume many digits until end of string.
 readNatural :: String -> Maybe Natural
 readNatural str
     | sz == 0  = Nothing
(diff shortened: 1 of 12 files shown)


tests/Checks.hs
```diff
@@ -2,6 +2,7 @@
 module Main where
```
ndmitchell: Is this deliberately exporting testAdditive? Otherwise `module Main(main) where` is probably desirable. I tend to always require export lists.

Foundation/String/UTF8.hs
```diff
@@ -1246,37 +1252,80 @@ readInteger :: String -> Maybe Integer
...
     | sz == 0  = Nothing
-    | otherwise = loop 0 0
+    | otherwise =
+        case decimalDigits 0 str 0 of
+            (# acc, True, endOfs #) | endOfs > 0 -> Just $ acc
+            _                                    -> Nothing
+  where
+    !sz = size str
+
+readFloatingExact :: String -> Maybe (Integer, Word, Natural)
```
ndmitchell: I'd recommend a bunch of comments on this. I've no idea what the Word represents, and no real idea how/what would be produced. I think these things work great with doc comments that are also tested.

vincenthz: The `FloatingExact` type is probably going away as an untyped tuple and will be represented by some new "primitive" types like `Ratio`. I'm not sure whether or not that's useful as is, or just an internal stepping stone to get to Double / Float

tests/Checks.hs
vincenthz: it's just the main module of the 2nd work-in-progress test-suite. I'll add the `(main)` but it make no practical difference I think

ndmitchell: Yep, it never makes any practical difference... Just something I was thinking about given I just added the check to HLint.

vincenthz:
Also as extra information related to this, I don't think this is the best method, but better methods are seriously more complicated. So the idea is to get something first, and then hopefully iterate with everything in place (tests, helpers ..)

PR #240, opened 2017-04-14
Add Ghc-8.2 in travis

vincenthz:
For now use rc1


.travis/stack-ghc-8.0.yaml
```diff
@@ -1,66 +1,6 @@
-# This file was automatically generated by 'stack init'
```
NicolasDP: much better indeed

NicolasDP:
Now it's a shame that all the links on haskell.org's download page are dead.

NicolasDP:
It seems the unavailability of the downloads.haskell.org was only temporary https://mail.haskell.org/pipermail/ghc-devs/2017-April/014124.html

vincenthz:
Probably need to be tweak to actually work, but since it's marked as allowed_failure, merging to move on and can revisit this soon

Issue #234, opened 2017-04-04
What is the best way to deal with stdin and stdout ?


Josef-Vonasek:
In classic Prelude I used to use interact whenever I could : 

```
main = interact $ unlines . map encodeToString . map myfunc . map decodeFromStrig . lines
```

In Foundation there is no interact nor unlines and lines. Should I write these functions myself or is there better way to deal with stdin and stdout? Is Parser meant for this?

vincenthz:
I would like to avoid all sort of lazy things, so here my thinking would be to have this workflow through conduit (completely untested and not even compiled check):

```haskell
import Foundation
import Foundation.Conduit
import Foundation.Conduit.Textual
import Foundation.String (Encoding(..))

main = runConduit $ sourceHandle stdin .| fromBytes UTF8 .| lines .| something myfunc .| toBytes UTF8 .|. sinkHandle stdout
```

ndmitchell:
`interact` was very sensitive to buffering and laziness, so it's never something I'd encourage. @vincenthz - in the above do you imagine `lines` is the `String` one or the conduit one?

vincenthz:
It would be conduit's `lines` one. In term of buffering I want to get rid of all the handle buffering when we move to foundation's `Handle`, which in essence would be the same as having `NoBuffering` on handle with no way to change.

I expect all the buffering could be better done through conduit in `sourceHandle` and `sinkHandle`, or possibly re-add some conduit like combinator to specify the sometimes useful `LineBuffering` and/or giving the ability to buffer differently.

vincenthz:
@ndmitchell for clarity, "never something I'd encourage" apply to base's interact right ? or to the rewrite using conduit ?

ndmitchell:
@vincenthz I'd never use interact. I've used it before (it was the only IO thing in my first Haskell course) and it's not something that should be repeated. 

I worry about no buffering - given a conduit the buffer strategy is obvious and if it's easy to screw up people will. I guess as long as handle to conduit buffers and does page sized reads all will be good.

vincenthz:
The buffering in Haskell (and in C with fopen/fread/fwrite family) is only to make something that shouldn't be done, faster; namely pretending that you're getting bytes/chars from the system at the speed the API want (e.g. hGetChar) instead of the speed of the underlaying device ("block size" for disk for example).

We have the luxury to make a much nicer api that was ever possible in C, through conduit (or people can use their things on the lowlevel IO api which are less magic)

[...]

PR #245, opened 2017-04-16
Conduit textual

vincenthz:
Add ability to deal with textual file/handle.

Also answer #234

### DIFF
--- a/Foundation/Conduit/Textual.hs
+++ b/Foundation/Conduit/Textual.hs
@@ -5,8 +5,9 @@ module Foundation.Conduit.Textual
     ) where
 
 import           Foundation.Internal.Base hiding (throw)
-import           Foundation.Array.Unboxed
+import           Foundation.Array.Unboxed (UArray)
 import           Foundation.String (String)
+import           Foundation.Collection
 import qualified Foundation.String.UTF8 as S
 import           Foundation.Conduit.Internal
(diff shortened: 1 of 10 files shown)


Foundation/Conduit/Textual.hs
```diff
@@ -0,0 +1,46 @@
...
+lines :: Monad m => Conduit String String m ()
+lines = await >>= maybe (finish mempty) (go mempty)
+  where
+    finish buf = if S.null buf then return () else yield buf
+
+    go current nextBuf =
+        case S.uncons next' of
+            Just (_, rest') -> yield (current `mappend` line) >> go mempty rest'
+            Nothing         ->
+                let nextCurrent = current `mappend` nextBuf
```
ndmitchell: This is going to be _O(n^2)_. I'd recommend a list of buffers and doing a reverse at the end (or a builder, if you want to be fancy). I had to submit a patch to fix a bug like this in the original conduit since it can easily come up (reading a huge file with no lines).

vincenthz: yeah good call; It did morally hurt me to actually merge the string this way, but had to be done in the spirit of getting something right now without more yak shaving. I'll seek to fix it before merging, but sadly the string builder is still pretty much only in my head

PR #247, opened 2017-04-17
Does not compile with GHC 7.4

alanz:


vincenthz:
it's not because of base that it doesn't compile, it's because of using unboxed tuples in pattern match that is not supported by the *compiler*.

alanz:
Yes, but I am using base as a proxy for the compiler, to rule out the older one that no longer works.

vincenthz:
That sounds semantically wrong (hypothetically: what if I have my own ghc 7.4 with support for unboxed tuple in pattern position but with the same base)

`Tested-with` is the compilers we support in foundation

ndmitchell:
Semantically wrong, yes. But still quite convenient... The package vs compiler distinction is broken, but the patch would still be valuable.

Issue #243, opened 2017-04-16
Inaccuracies introduced by readDouble


ndmitchell:
Given:

```haskell
print $ readDouble $ fromList "-31.548"
print $ (read "-31.548" :: Double)
```

I get:

```
Just (-31.548000000000002)
-31.548
```

The foundation read is slightly off.

vincenthz:
This is very linked to the way it's built from integral + floating ("simpler" algorithm) instead of in one go (which is the optimised way I'm aluding to in the other ticket). `Data.Text.Read.double` has the same construction and give the same approximation.

ndmitchell:
Ah, I didn't realise readDouble had basically identical issues. I guess I'm just spoilt by the standard `read/show` being so robust for `Double`.

PR #248, opened 2017-04-17
add bench for reading & optimise

vincenthz:
After correcting the initial wrong benchmark, we're slighlty slower in all cases so far (not unexpected):

Integer:

```
benchmarking types/String/Read/Integer/10000/String
time                 200.1 ns   (197.5 ns .. 202.9 ns)
benchmarking types/String/Read/Integer/10000/Text
time                 126.0 ns   (124.1 ns .. 128.4 ns)

benchmarking types/String/Read/Integer/1234567891234567890/String
time                 798.4 ns   (786.5 ns .. 809.1 ns)
benchmarking types/String/Read/Integer/1234567891234567890/Text
time                 505.7 ns   (497.8 ns .. 513.1 ns)
```

Double:

```
benchmarking types/String/Read/Double/100.56e23/String
time                 590.6 ns   (584.3 ns .. 597.0 ns)
benchmarking types/String/Read/Double/100.56e23/Text
time                 323.7 ns   (318.4 ns .. 329.3 ns)
benchmarking types/String/Read/Double/-123.1247/String
time                 597.0 ns   (589.5 ns .. 603.8 ns)
benchmarking types/String/Read/Double/-123.1247/Text
time                 315.6 ns   (308.9 ns .. 321.6 ns)
```

### DIFF
--- a/benchs/Fake/Text.hs
+++ b/benchs/Fake/Text.hs
@@ -7,6 +7,7 @@ module Fake.Text
     , any
     , filter
     , reverse
+    , decimal
     ) where
 
 import Prelude (undefined)
@@ -20,3 +21,4 @@ take        = undefined
 filter _    = undefined
 reverse     = undefined
 any         = undefined
+decimal     = undefined
--- a/benchs/Fake/Text.hs
+++ b/benchs/Fake/Text.hs
@@ -11,7 +11,7 @@ module Fake.Text
     , double
     ) where
 
-import Prelude (undefined)
+import Prelude (undefined, Either(..))
 
 data Text = Text
 
@@ -22,5 +22,9 @@ take        = undefined
 filter _    = undefined
 reverse     = undefined
 any         = undefined
-decimal     = undefined
-double      = undefined
+
+decimal :: Text -> Either a (b, c)
+decimal = undefined
+
+double :: Text -> Either a (b, c)
+double = undefined
--- a/benchs/Fake/Text.hs
+++ b/benchs/Fake/Text.hs
@@ -8,6 +8,7 @@ module Fake.Text
     , filter
     , reverse
     , decimal
+    , double
     ) where
 
 import Prelude (undefined)
@@ -22,3 +23,4 @@ filter _    = undefined
 reverse     = undefined
 any         = undefined
 decimal     = undefined
+double      = undefined
--- a/Foundation/Check.hs
+++ b/Foundation/Check.hs
@@ -112,7 +112,7 @@ runProp ctx s prop = iterProp 1
         loop :: Word -> [String] -> [String]
         loop _ []      = printChecks checks
         loop !i (a:as) = "parameter " <> fromList (show i) <> " : " <> a <> "\n" : loop (i+1) as
-    printChecks (PropertyBinaryOp True name _ _)  = []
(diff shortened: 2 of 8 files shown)


benchs/Fake/Text.hs
```diff
@@ -8,6 +8,7 @@ module Fake.Text
     , filter
     , reverse
     , decimal
+    , double
```
ndmitchell: I don't really get the purpose of Fake.Text - perhaps it needs to say at the top of the module why it might be useful? Or perhaps because its in benchs that's not a problem.

vincenthz: It's just a contraption to allow building without external dependencies like text, bytestring, etc.. when doing internal benchs (instead of comparative benchs with other libraries).
The fake module limits the number of ifdef in the bench modules.

ndmitchell: Ah, that makes a lot of sense. Thanks.

vincenthz:
Integer:
```
benchmarking types/String/Read/Integer/10000/String
time                 116.5 ns   (115.2 ns .. 118.0 ns)

benchmarking types/String/Read/Integer/10000/Text
time                 115.6 ns   (114.4 ns .. 117.2 ns)

benchmarking types/String/Read/Integer/1234567891234567890/String
time                 462.6 ns   (453.5 ns .. 471.1 ns)

benchmarking types/String/Read/Integer/1234567891234567890/Text
time                 489.0 ns   (479.0 ns .. 496.2 ns)
```

Int:

```
benchmarking types/String/Read/Int/12345/String
time                 34.93 ns   (34.31 ns .. 35.61 ns)

benchmarking types/String/Read/Int/12345/Text
time                 36.91 ns   (36.31 ns .. 37.51 ns)

benchmarking types/String/Read/Int/12345/ByteString
time                 28.36 ns   (27.71 ns .. 29.15 ns)
```

vincenthz:
55f655b96986a46d822185f946caf45596178c59 make readDouble less imprecise and fix #243

vincenthz:
So now, foundation is more precise and faster than Text to parse Double.

```
benchmarking types/String/Read/Double/100.56e23/String
time                 246.4 ns   (242.7 ns .. 250.6 ns)
             
benchmarking types/String/Read/Double/100.56e23/Text
time                 315.3 ns   (310.2 ns .. 319.9 ns)
             
benchmarking types/String/Read/Double/-123.1247/String
time                 223.4 ns   (220.1 ns .. 226.2 ns)
             
benchmarking types/String/Read/Double/-123.1247/Text
time                 308.8 ns   (304.1 ns .. 313.1 ns)
```

vincenthz:
For the precision thing, this is always going to be a tradeoff. It would be more precise to work out the encoding using `fromRat''` without a `/` operation, but it would also be very likely slower (and also approximate nonetheless). In practice, I haven't work out how much imprecision that add, but that would be surprising it that was that significative.

What about adding a readRational that keeps the precision, and then the user can manually call `fromRational` which would brings the same algorithm used by `read @ Double` ?

ndmitchell:
readRational seems like an excellent idea.

vincenthz:
it's all in now. probably could do more optimisation, but that's good enough for this iteration I think.

Issue #249, opened 2017-04-17
Does not build on RHEL 5.8


mlongob:
```
$ cat /etc/redhat-release
Red Hat Enterprise Linux Server release 5.8 (Tikanga)
```

### Build error in `Posix.hsc`
```
/.../foundation/In file included from .stack-work/dist/x86_64-linux/Cabal-1.24.2.0/build/Foundation/System/Bindings/Posix_hsc_make.c:1:0:
Posix.hsc: In function ‘main’:
/.../foundation/Posix.hsc:243:16: error: ‘O_NOFOLLOW’ undeclared (first use in this function)
ghc/template-hsc.h:35:10: note: in definition of macro ‘hsc_const’
     if ((x) < 0)                                      \
          ^
/.../foundation/Posix.hsc:243:16: note: each undeclared identifier is reported only once for each function it appears in
ghc/template-hsc.h:35:10: note: in definition of macro ‘hsc_const’
     if ((x) < 0)                                      \
          ^
/.../foundation/Posix.hsc:244:16: error: ‘O_CLOEXEC’ undeclared (first use in this function)
ghc/template-hsc.h:35:10: note: in definition of macro ‘hsc_const’
     if ((x) < 0)                                      \
```

### Build error in `Linux.hsc`
```
/.../foundation/In file included from .stack-work/dist/x86_64-linux/Cabal-1.24.2.0/build/Foundation/System/Bindings/Linux_hsc_make.c:1:0:
Linux.hsc: In function ‘main’:
/.../foundation/Linux.hsc:43:16: error: ‘IN_NONBLOCK’ undeclared (first use in this function)
ghc/template-hsc.h:35:10: note: in definition of macro ‘hsc_const’
     if ((x) < 0)                                      \
          ^
/.../foundation/Linux.hsc:43:16: note: each undeclared identifier is reported only once for each function it appears in
ghc/template-hsc.h:35:10: note: in definition of macro ‘hsc_const’
     if ((x) < 0)                                      \
          ^
/.../foundation/Linux.hsc:44:16: error: ‘IN_CLOEXEC’ undeclared (first use in this function)
ghc/template-hsc.h:35:10: note: in definition of macro ‘hsc_const’
     if ((x) < 0)
```

### Build error in `foundation_random.c`
```
/.../foundation/In file included from cbits/foundation_random.c:12:0: error:

/usr/include/linux/random.h:38:2: error:
     error: unknown type name ‘__u32’
      __u32 buf[0];
[...]

PR #250, opened 2017-04-17
Build compatibility fixes for RHEL 5.8

mlongob:
Fixes build issues in #249


vincenthz:
RHEL 5.8 .. is really old !

vincenthz:
Thanks for the PR !

mlongob:
Really old indeed. Still using it somewhere though... Thanks for the quick merge

PR #254, opened 2017-04-20
add a class MonadState and add support for StateT

### LINKED ISSUE #246: Is Foundation.Monad.State usable?
I'm looking at the `StateT` type with some confusion. Is there no way to `put` the state?

NicolasDP:
fix #246


vincenthz:
It's failing for old versions where functor/applicative was not a parent of monad

PR #256, opened 2017-04-21
add support for DList

NicolasDP:
# DList

I have implemented a difference list. I think it might be useful for the `Builder` and might become useful for the other algorithm that require appending element at `O(1)` cost.

# Yak shaving

Sorry this PR does more than just DList, I could not resist starting to port a couple of Quickcheck tests to our great Foundation's Check.


ndmitchell:
DList is a particular encoding, but I've generally found that `data Str a = Zero | One a | Two (Str a) (Str a)` often performs better, especially if you add another constructor for tail-append (e.g. `Two1 (Str a) a`). You can then traverse that to build a list (or anything else) quickly. Have you benchmarked DList against such a structure? Certainly my knowledge is quite old, so compiler optimisations may have made one or the other much cheaper since then.

Foundation/List/DList.hs
```diff
@@ -0,0 +1,89 @@
...
+    fmap f = foldr (cons . f) mempty
+
+instance Applicative DList where
+    pure = singleton
+    (<*>) m1 m2 = m1 >>= \x1 -> m2 >>= \x2 -> return (x1 x2)
+
+instance Monad DList where
+    (>>=) m k = foldr (mappend . k) mempty m
+    return = singleton
+    fail _ = mempty
```
vincenthz: we don't need fail, it's going to be in MonadFail, and DList won't have the instance anyway

Foundation/List/DList.hs
```diff
@@ -0,0 +1,89 @@
...
+    foldr f b = foldr f b . toList
+    foldl f b = foldl f b . toList
+    foldl' f b = foldl' f b . toList
+
+instance Collection (DList a) where
+    null = null . toList
+    length = length . toList
+    elem a = elem a . toList
+    maximum = maximum . nonEmpty_ . toList
+    minimum = minimum . nonEmpty_ . toList
```
vincenthz: thinking out loud: it would be nice to have a toNonEmptyList that does `nonEmpty_ . toList` in one place with the right strong type.

vincenthz:
@ndmitchell that sounds almost like some kind of degenerate tree / zipper. Possibly quite interesting, although on the list of yak shaving I'll put it quite low

vincenthz:
@NicolasDP good work on the tests ! I see the properties are already in a library form, so that will be useful eventually

NicolasDP:
@ndmitchell I didn't consider implementing this other list. It is more straight forward and would certainly needs less `toList` in the instantiation of the different Collection's classes.

I see that @vincenthz has already created a ticket to implement your proposal. I guess we can merge this one and see later for a better encoding of the `DList`.

Foundation/List/DList.hs
NicolasDP: make sense. removing it.

PR #260, opened 2017-04-24
Fix broken link

teodorlu:


vincenthz:
Thanks. as soon as I get the readthedoc bits working, this is very likely going to change again

PR #266, opened 2017-04-27
Block & MutableBlock

vincenthz:
This is a leaner version of `UArray` which only works on top of the `Bytearray#` and doesn't have any slicing ability (no builtin size/offset). It's also not C friendly by default.

People shouldn't be using this generally but in certain case, using a Block inside a data structure to represent some binary data have significant advantages:

* No meta data: saving ~20 bytes or so compared to a `UArray`.
* No "part" sharing: if you set a field of a data structure to some `UArray`, specially after a take/drop, you're still holding on to the bigger UArray. so "having a Block" is more straightforward in memory tracking.
* Unpackable in a parent: since it doesn't have a sum type like UArray

possibly this could bring some further cleanup & refactoring in `UArray`

### DIFF
--- a/Foundation/Array.hs
+++ b/Foundation/Array.hs
@@ -24,7 +24,7 @@ module Foundation.Array
     , OutOfBound
     ) where
 
-import           Foundation.Array.Common
+import           Foundation.Primitive.Exception
 import           Foundation.Array.Boxed
 import           Foundation.Array.Unboxed
 import           Foundation.Array.Unboxed.Mutable
--- a/Foundation/Array/Bitmap.hs
+++ b/Foundation/Array/Bitmap.hs
@@ -32,7 +32,7 @@ module Foundation.Array.Bitmap
 import           Foundation.Array.Unboxed (UArray)
 import qualified Foundation.Array.Unboxed as A
 import           Foundation.Array.Unboxed.Mutable (MUArray)
-import           Foundation.Array.Common
+import           Foundation.Primitive.Exception
 import           Foundation.Internal.Base
 import           Foundation.Primitive.Types.OffsetSize
 import           Foundation.Primitive.Monad
--- a/Foundation/Array/Boxed.hs
+++ b/Foundation/Array/Boxed.hs
@@ -74,7 +74,7 @@ import           Foundation.Primitive.Types
 import           Foundation.Primitive.NormalForm
 import           Foundation.Primitive.IntegralConv
 import           Foundation.Primitive.Monad
-import           Foundation.Array.Common
+import           Foundation.Primitive.Exception
 import           Foundation.Boot.Builder
 import qualified Foundation.Boot.List as List
 import qualified Prelude
--- a/Foundation/Array/Chunked/Unboxed.hs
+++ b/Foundation/Array/Chunked/Unboxed.hs
@@ -21,7 +21,7 @@ import           Data.Typeable
 import           Control.Arrow ((***))
 import           Foundation.Array.Boxed (Array)
 import qualified Foundation.Array.Boxed as A
-import           Foundation.Array.Common
+import           Foundation.Primitive.Exception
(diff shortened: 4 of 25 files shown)


ndmitchell:
Do you intend to use Block in UVecBA?

vincenthz:
@ndmitchell it's very likely I will try to, that's the intent behind the "cleanup/refactor of the UArray" that I mentioned. If it doesn't yield about performance nasties, it would simplify and cleanup quite a few code path in UArray.

vincenthz:
Adding a small TODO to keep track:

* [ ] Add missing functions (take, drop,..)
* [ ] Add the necessary type classes (Show,Eq,Ord,..)
* [ ] Add the foundation type classes (Hashable, Collection, Sequential, NormalForm)
* [ ] Add tests

ndmitchell:
Running this branch on the ToForeign benchmark, I get:

```
benchmarking toByteString-copyPtr
Warning: Couldn't open /dev/urandom
Warning: using system clock for seed instead (quality will be lower)
time                 30.82 ns   (30.74 ns .. 30.92 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 30.81 ns   (30.73 ns .. 30.95 ns)
std dev              355.7 ps   (254.4 ps .. 562.7 ps)
variance introduced by outliers: 12% (moderately inflated)

benchmarking toByteString-withPtr
time                 107.8 ns   (106.4 ns .. 109.6 ns)
                     0.999 R²   (0.998 R² .. 0.999 R²)
mean                 107.7 ns   (106.9 ns .. 108.7 ns)
std dev              2.958 ns   (2.498 ns .. 3.613 ns)
variance introduced by outliers: 42% (moderately inflated)

benchmarking toByteString-native-copyPtr
time                 29.36 ns   (29.29 ns .. 29.43 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 29.38 ns   (29.30 ns .. 29.47 ns)
std dev              290.2 ps   (205.6 ps .. 443.7 ps)

benchmarking toByteString-native-withPtr
time                 143.7 ns   (143.3 ns .. 144.2 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 143.8 ns   (143.4 ns .. 144.6 ns)
std dev              1.869 ns   (1.324 ns .. 2.998 ns)
variance introduced by outliers: 13% (moderately inflated)

benchmarking toByteString-block-copyPtr
time                 26.16 ns   (26.10 ns .. 26.23 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 26.19 ns   (26.12 ns .. 26.30 ns)
std dev              301.8 ps   (205.0 ps .. 516.0 ps)
variance introduced by outliers: 12% (moderately inflated)
```

If I repeat my trick of `INLINE` on `withPtr` that drops to:

```
benchmarking toByteString-withPtr
time                 31.07 ns   (30.90 ns .. 31.39 ns)
                     0.998 R²   (0.995 R² .. 1.000 R²)
mean                 32.06 ns   (31.56 ns .. 33.30 ns)
std dev              2.392 ns   (1.112 ns .. 4.861 ns)
variance introduced by outliers: 85% (severely inflated)

benchmarking toByteString-native-withPtr
time                 43.24 ns   (43.12 ns .. 43.34 ns)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 43.23 ns   (43.08 ns .. 43.38 ns)
std dev              526.0 ps   (429.8 ps .. 702.1 ps)
variance introduced by outliers: 13% (moderately inflated)
```

In the new branch I don't actually want to use withPtr, so I'm not as worried, but it seems that it's generally useful and should be available and cheapish?

vincenthz:
thanks for re-trying the bench, I've forgot to push the INLINE of withPtr when going through pushing commits; the pragma is unfortunate but I agree it's too substantial to ignore.

PR #272, opened 2017-04-28
add sum type These

NicolasDP:
I think this can come in handy sooner or later.


vincenthz:
:+1: however I think adding just a type export in one single exposed module is not scalable. Probably should go in primitive and exported as part of all the "normal basic stuff" that foundation exposes.

PR #273, opened 2017-04-28
Check: parameters and other output improvements

### LINKED ISSUE #270: [Check] Add basic command line argument parsing for seed setting & pattern matching tests want to run
At minimum:

```
--seed <seed-value> : set the seed to a specific value
-p <pattern-match> : only run tests that somehow (method to be defined) match the name of test
-l : list-all-tests
```

NicolasDP:
also fix #270 

This is still work in progress I think. But is providing the basic feature we need (parameterising `--seed` and configuring verbosity of the output).

### DIFF
--- a/Foundation/Check.hs
+++ b/Foundation/Check.hs
@@ -59,6 +59,11 @@ testName (Unit s _)     = s
 testName (Property s _) = s
 testName (Group s _)    = s
 
+groupHasSubGroup :: [Test] -> Bool
+groupHasSubGroup [] = False
+groupHasSubGroup (Group{}:_) = True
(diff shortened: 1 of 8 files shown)


vincenthz:
The problem with unicode and windows:

```       check-foundation.exe: <stdout>: commitBuffer: invalid argument (invalid character)
```

vincenthz:
And missing the import of Proxy from Foundation.Internal.Proxy

vincenthz:
I'm not sure about the improvement to the display bits, but this is a clear improvement for argument parsing and towards a more complete checks. thanks !

PR #282, opened 2017-05-09
Add foundation Semigroup&Monoid

vincenthz:
The reason to add:

* base Semigroup currently use NonEmpty which is different in foundation
* base Semigroup use an Int as parameter for stimes however negative
  and zero values are wrong and not prevented by design.
* base Monoid doesn't depends on Semigroup

The caveat:

* in practice that means interacting with base, and regular Semigroup & Monoid stuff
  is harder.

### DIFF
--- a/Foundation/Boot/List.hs
+++ b/Foundation/Boot/List.hs
@@ -1,6 +1,7 @@
 module Foundation.Boot.List
     ( length
     , sum
+    , foldr
     ) where
 
 import Foundation.Internal.Base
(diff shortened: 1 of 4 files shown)


ThreeFx:
Should `Foundation.Class.Semigroup` also export `Sum`, `Product`, `Max`, etc. newtypes? We probably want to reexport them from `Foundation.Class.Monoid` too.

vincenthz:
if foundation does end up exporting its own Semigroup and Monoid, then yeah I would probably expect that we'll have to export those types as foundation-friendly version.

PR #283, opened 2017-05-10
Fix up foundation.cabal to obey "neil check" invariants

ndmitchell:
Mostly casing of the cabal fields, but there are a couple of other little things (year must be in the copyright field, must use extra-docs)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2015-2016 Vincent Hanquez <vincent@snarc.org>
+Copyright (c) 2015-2016 Vincent Hanquez <vincent@snarc.org> 2016-2017
 
 All rights reserved.
 
--- a/foundation.cabal
+++ b/foundation.cabal
@@ -1,7 +1,7 @@
-Name:                foundation
-Version:             0.0.9
-Synopsis:            Alternative prelude with batteries and no dependencies
-Description:
+name:                foundation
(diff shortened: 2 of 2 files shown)


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2015-2016 Vincent Hanquez <vincent@snarc.org>
+Copyright (c) 2015-2016 Vincent Hanquez <vincent@snarc.org> 2016-2017
```
vincenthz: that's a weird format, usually the year is before the name and after the diff it's on both side. It would probably make sense to just drop the years, that hasn't been required for many many years now ..

ndmitchell: I think the complaint was that 2017 wasn't in the copyright year, so making it 2015-2017 would fix neil check anyway. However, if you think dropping the year is preferable and not required, do you have a link that says/explains that? Given the `neil check` tool is fairly strict and updating stuff once a year is a pain I'd be happy to get rid of the date entirely.

vincenthz: I don't really have a link for that, I derive this information from the legal department of a previous big company TM. Most things I'm finding on the internet seems either confused, outdated, or just claim something different, so sadly no certainty here.

copyright is valid regardless of the copyright notice anyway, it's just easier to claim "accidental violation of copyright" if the code is not littered with copyright notice.

PR #285, opened 2017-05-17
Misc changes

### LINKED ISSUE #284: Add copyFromPtr
```
newFromPtr :: Count -> Ptr ty -> prim (UArray ty)
```

Which copies the array from a Ptr, into either a pinned or unpinned lump as appropriate. I couldn't see any way to do that equally efficiently outside the library.

Also possible to do `copyFromPtr` as well?

vincenthz:
* add --list-tests to checks
* lowlevel optimisation with size/Offset casting
* add `copyFromPtr` and `freezeFromPtr` fix #284 
* speed up `compare` and `==` for string, uarray and block


vincenthz:
Also fixes #286

PR #296, opened 2017-06-09
Initial HLint work

galderz:
Initial HLint work fixing warnings.

Error free `hlint` output can be achieved via:

```
hlint --cpp-define=__GLASGOW_HASKELL__=800 --cpp-define=x86_64_HOST_ARCH=1 --cpp-define=mingw32_HOST_OS=1 .
```

Still TODO:
- [x] Add documentation on HLint.
- [x] Move "How To Contribute" section to docs section and add HLint there. 
- [ ] Link from README to contributing guide.
- [x] Deal with suggestions.
- [ ] Add to CI.

### DIFF
--- a/Foundation/Array/Unboxed.hs
+++ b/Foundation/Array/Unboxed.hs
@@ -143,8 +143,8 @@ arrayType :: DataType
 arrayType = mkNoRepType "Foundation.UArray"
 
 instance NormalForm (UArray ty) where
-    toNormalForm (UVecBA _ _ _ _) = ()
-    toNormalForm (UVecAddr _ _ _) = ()
+    toNormalForm (UVecBA _ _ _ !_) = ()
+    toNormalForm (UVecAddr {}) = ()
 instance (PrimType ty, Show ty) => Show (UArray ty) where
     show v = show (toList v)
 instance (PrimType ty, Eq ty) => Eq (UArray ty) where
--- a/Foundation/Array/Unboxed.hs
+++ b/Foundation/Array/Unboxed.hs
@@ -1019,7 +1019,7 @@ reverse a
         loop !i
             | i == end  = return ()
             | otherwise = primMbaWrite ma i (primBaIndex ba (sizeAsOffset (endI - i))) >> loop (i+Offset 1)
-    goAddr !end !ma !(Ptr ba) !srcStart = loop (Offset 0)
+    goAddr !end !ma (Ptr ba) !srcStart = loop (Offset 0)
       where
         !endI = sizeAsOffset ((srcStart + end) - Offset 1)
         loop !i
--- a/Foundation/Array/Unboxed.hs
+++ b/Foundation/Array/Unboxed.hs
@@ -12,7 +12,6 @@
 --
 {-# LANGUAGE MagicHash #-}
 {-# LANGUAGE UnboxedTuples #-}
-{-# LANGUAGE ViewPatterns #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE Rank2Types #-}
 module Foundation.Array.Unboxed
--- a/Foundation/Array/Unboxed/Mutable.hs
+++ b/Foundation/Array/Unboxed/Mutable.hs
@@ -163,8 +163,8 @@ new sz
 mutableSame :: MUArray ty st -> MUArray ty st -> Bool
 mutableSame (MUVecMA sa ea _ ma) (MUVecMA sb eb _ mb) = and [ sa == sb, ea == eb, bool# (sameMutableByteArray# ma mb)]
 mutableSame (MUVecAddr s1 e1 f1) (MUVecAddr s2 e2 f2) = and [ s1 == s2, e1 == e2, finalPtrSameMemory f1 f2 ]
-mutableSame (MUVecMA {})     (MUVecAddr {})   = False
-mutableSame (MUVecAddr {})   (MUVecMA {})     = False
+mutableSame MUVecMA {}     MUVecAddr {}   = False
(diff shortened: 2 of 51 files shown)


galderz:
I've just updated the PR to fix the error that CI had found, add more ignores and finally add suggestions. Assuming CI passes, only docu would be left to do.

Foundation/Monad.hs
```diff
@@ -1,7 +1,6 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DeriveDataTypeable #-}
-{-# LANGUAGE StandaloneDeriving #-}
```
NicolasDP: I think this one is needed for older compilers (ghc.7.8 and 7.10).

see: https://travis-ci.org/haskell-foundation/foundation/jobs/241461362#L300

galderz: Yup, CI spotted it. Just fixed it.

Foundation/Array/Unboxed.hs
```diff
@@ -1020,7 +1019,7 @@ reverse a
         loop !i
             | i == end  = return ()
             | otherwise = primMbaWrite ma i (primBaIndex ba (sizeAsOffset (endI - i))) >> loop (i+Offset 1)
-    goAddr !end !ma !(Ptr ba) !srcStart = loop (Offset 0)
+    goAddr !end !ma (Ptr ba) !srcStart = loop (Offset 0)
```
ndmitchell: Should this have the ! on `ba`?

Foundation/Array/Unboxed/Mutable.hs
```diff
@@ -161,18 +161,18 @@ new sz
 {-# INLINE new #-}
 
 mutableSame :: MUArray ty st -> MUArray ty st -> Bool
-mutableSame (MUVecMA sa ea _ ma) (MUVecMA sb eb _ mb) = and [ sa == sb, ea == eb, bool# (sameMutableByteArray# ma mb)]
-mutableSame (MUVecAddr s1 e1 f1) (MUVecAddr s2 e2 f2) = and [ s1 == s2, e1 == e2, finalPtrSameMemory f1 f2 ]
-mutableSame (MUVecMA {})     (MUVecAddr {})   = False
-mutableSame (MUVecAddr {})   (MUVecMA {})     = False
+mutableSame (MUVecMA sa ea _ ma) (MUVecMA sb eb _ mb) = (sa == sb) && (ea == eb) && bool# (sameMutableByteArray# ma mb)
```
ndmitchell: I wouldn't usually bother with the brackets around `sa == sb`

Foundation/Check/Main.hs
```diff
@@ -67,11 +67,11 @@ filterTestMatching cfg testRoot
...
     | otherwise                = testFilter [] testRoot
   where
-    match acc s = or $ fmap (flip isInfixOf currentTestName) $ testNameMatch cfg
+    match acc s = or (flip isInfixOf currentTestName <$> testNameMatch cfg)
       where currentTestName = fqTestName (s:acc)
     or [] = False
     or (x:xs)
-        | x == True = True
+        | x         = True
         | otherwise = or xs
```
ndmitchell: Isn't this just absolutely standard `or` from the base library? Why replicate it?

ndmitchell:
This should probably be enabled on the CI as well - see https://github.com/ndmitchell/hlint#running-with-continuous-integration

galderz:
> This should probably be enabled on the CI as well - see https://github.com/ndmitchell/hlint#running-with-continuous-integration

Certainly. I'll update the TODO list at the top.

Foundation/Array/Unboxed.hs
galderz: Hmmmm, dunno... @vincenthz ?

Foundation/Array/Unboxed/Mutable.hs
galderz: Ah, true... From chat w/ @vincenthz yday, at times he likes having parenthesis to read the code easier. I'll let him decide on this one.

NicolasDP:
Good idea about adding hlint in our CI too. About Travis, I would rather like we only run the hlint in a different job. A job dedicated for running hlint.

ndmitchell:
How come a dedicated job? Should only add a few seconds to the time. When you add weeder (I hope!) would you also put that in a dedicated job?

galderz:
Yeah, I'll check with the guys later on how to add a job that runs first with just `HLint`.

Indeed, planning to add Weeder too. I'll create a separate issue for this.

galderz:
Btw, I'm keeping an eye on build results, fixing stuff and pushing new changes... that's why CI keeps rebuilding 😎

NicolasDP:
I think it would make things easier to spot on travis-ci.

* one job per compiler;
* one job for hlint;
* one job for weeder and detecting dead code ? Why not...
* one job for doctest ?

The idea is to be able to quickly see what the problems are by just looking at the job list for the associated PR. Avoiding to spend too much time scrolling the logs to find that it's not the compilation or the tests who failed but a linting issue or a documentation issue...

This is open to discussion anyway. I'm not going to make the decision for the whole community.

ndmitchell:
That kind of makes sense - apart from doctest - you really want to run the doctests in every compiler since they might operate differently. Things like HLint and Weeder I can see the reasoning.

galderz:
Added another commit with documentation and moved the contributing guide to docs.

Foundation/Check/Main.hs
galderz: I've created #316

PR #297, opened 2017-06-09
Rewrite 'uncurryN' to use 'Fstable' etc., add 'curryN'

ThreeFx:
Moved `uncurry` from `Foundation.Collection.Zippable` to `Foundation.Tuple`.
 - Reason: `uncurry` is a more powerful abstraction and useful as a `Foundation.Tuple` export.

Implemented `curry` in `Foundation.Tuple`. Also export `curry`.

`uncurry2` and `uncurry3` now use `Fstable`, `Sndable` and `Thdable`.

Further ideas:
 - Make any instance of `Fstable` also an instance of `Nthable 1` (e.g. `instance Fstable a => Nthable 1 a`).
 - Same for `Sndable`, etc.

### DIFF
--- a/Foundation/Collection/Zippable.hs
+++ b/Foundation/Collection/Zippable.hs
@@ -21,8 +21,9 @@ import qualified Foundation.Array.Boxed as BA
 import qualified Foundation.String.UTF8 as S
 import           Foundation.Collection.Element
 import           Foundation.Collection.Sequential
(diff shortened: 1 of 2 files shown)


vincenthz:
so I'm just going to close this for now. I think we all agree that it wasn't necessary such a good idea, thanks for both trying in any case. also happy to revisit any part in the future ..

Issue #299, opened 2017-06-09
maximum & minimum is implemented through a list


vincenthz:
Implementation of maximum & minimum is not optimal at all and used a list for most types:

```haskell
instance Collection (BA.Array ty) where
    ...
    minimum = Data.List.minimum . toList . getNonEmpty -- TODO
    ...
```

types having a bad implementation:

* [ ] Block
* [ ] UArray
* [ ] Array
* [ ] String
* [ ] Bitmap
* [ ] ChunkedUArray

PR #313, opened 2017-06-10
Optimize `Collection' functions for internal types

ThreeFx:
I chose to go with a `foldl'` implementation of these functions.

Also, `ChunkUArray` can be made an instance of `Foldable` by double-folding on the arrays, which allows the `Collection` functions to utilise those `fold`s.

### DIFF
--- a/Foundation/Array/Bitmap.hs
+++ b/Foundation/Array/Bitmap.hs
@@ -91,8 +91,8 @@ instance C.Collection Bitmap where
                  in foldl' min (unsafeIndex bm' 0) bm'
     maximum bm = let bm' = C.getNonEmpty bm
                  in foldl' max (unsafeIndex bm' 0) bm'
-    all p      = foldl' (\acc x -> p x && acc) True
-    any p      = foldl' (\acc x -> p x || acc) False
+    all = all
+    any = any
 
 instance C.Sequential Bitmap where
     take = take
@@ -445,6 +445,24 @@ foldl' f initialAcc vec = loop 0 initialAcc
     loop i !acc
         | i .==# len = acc
         | otherwise  = loop (i+1) (f acc (unsafeIndex vec i))
+
+all :: (Bool -> Bool) -> Bitmap -> Bool
+all p bm = loop 0
+  where
+    len = length bm
+    loop !i
+      | i .==# len = True
+      | not $ p (unsafeIndex bm i) = False
+      | otherwise = loop (i + 1)
+
+any :: (Bool -> Bool) -> Bitmap -> Bool
+any p bm = loop 0
+  where
+    len = length bm
+    loop !i
+      | i .==# len = False
+      | p (unsafeIndex bm i) = True
+      | otherwise = loop (i + 1)
 
 unoptimised :: ([Bool] -> [Bool]) -> Bitmap -> Bitmap
 unoptimised f = vFromList . f . vToList
--- a/Foundation/Array/Bitmap.hs
+++ b/Foundation/Array/Bitmap.hs
@@ -87,10 +87,13 @@ instance C.Collection Bitmap where
     null = null
     length = length
     elem e = Data.List.elem e . toList
-    minimum = Data.List.minimum . toList . C.getNonEmpty -- TODO can shortcircuit all this massively
-    maximum = Data.List.maximum . toList . C.getNonEmpty -- TODO DITTO
-    all p = Data.List.all p . toList
-    any p = Data.List.any p . toList
+    minimum bm = let bm' = getNonEmpty bm
+                 in foldl' min (UV.unsafeIndex bm' 0) bm'
+    maximum bm = let bm' = getNonEmpty bm
+                 in foldl' max (UV.unsafeIndex bm' 0) bm'
+    all p      = foldl' (\acc x -> p x && acc) True
+    any p      = foldl' (\acc x -> p x || acc) False
+
 instance C.Sequential Bitmap where
     take = take
     drop = drop
--- a/Foundation/Array/Bitmap.hs
+++ b/Foundation/Array/Bitmap.hs
@@ -87,10 +87,10 @@ instance C.Collection Bitmap where
     null = null
     length = length
     elem e = Data.List.elem e . toList
-    minimum bm = let bm' = getNonEmpty bm
-                 in foldl' min (UV.unsafeIndex bm' 0) bm'
-    maximum bm = let bm' = getNonEmpty bm
-                 in foldl' max (UV.unsafeIndex bm' 0) bm'
+    minimum bm = let bm' = C.getNonEmpty bm
+                 in foldl' min (unsafeIndex bm' 0) bm'
+    maximum bm = let bm' = C.getNonEmpty bm
+                 in foldl' max (unsafeIndex bm' 0) bm'
     all p      = foldl' (\acc x -> p x && acc) True
     any p      = foldl' (\acc x -> p x || acc) False
 
--- a/Foundation/Array/Bitmap.hs
+++ b/Foundation/Array/Bitmap.hs
@@ -87,10 +87,8 @@ instance C.Collection Bitmap where
     null = null
     length = length
     elem e = Data.List.elem e . toList
-    minimum bm = let bm' = C.getNonEmpty bm
-                 in foldl' min (unsafeIndex bm' 0) bm'
-    maximum bm = let bm' = C.getNonEmpty bm
-                 in foldl' max (unsafeIndex bm' 0) bm'
+    maximum = any id . C.getNonEmpty
+    minimum = all id . C.getNonEmpty
     all = all
     any = any
 
--- a/Foundation/Array/Chunked/Unboxed.hs
+++ b/Foundation/Array/Chunked/Unboxed.hs
@@ -67,8 +67,8 @@ instance PrimType ty => C.Collection (ChunkedUArray ty) where
     elem   = elem
     minimum = minimum
     maximum = maximum
-    all p = foldl' (\acc x -> p x && acc) True
-    any p = foldl' (\acc x -> p x || acc) False
+    all p (ChunkedUArray cua) = A.all (U.all p) cua
+    any p (ChunkedUArray cua) = A.any (U.any p) cua
 
 instance PrimType ty => C.Sequential (ChunkedUArray ty) where
     take = take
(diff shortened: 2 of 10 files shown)


ThreeFx:
Oh, this also fixes #299.

Foundation/Array/Bitmap.hs
```diff
@@ -87,10 +87,13 @@ instance C.Collection Bitmap where
     null = null
     length = length
     elem e = Data.List.elem e . toList
-    minimum = Data.List.minimum . toList . C.getNonEmpty -- TODO can shortcircuit all this massively
-    maximum = Data.List.maximum . toList . C.getNonEmpty -- TODO DITTO
-    all p = Data.List.all p . toList
-    any p = Data.List.any p . toList
+    minimum bm = let bm' = C.getNonEmpty bm
+                 in foldl' min (unsafeIndex bm' 0) bm'
```
ndmitchell: Does this end up applying `min` twice on the 0 index? Should you be using foldl1'?

Foundation/Array/Bitmap.hs
```diff
@@ -87,10 +87,13 @@ instance C.Collection Bitmap where
...
+    maximum bm = let bm' = C.getNonEmpty bm
+                 in foldl' max (unsafeIndex bm' 0) bm'
+    all p      = foldl' (\acc x -> p x && acc) True
+    any p      = foldl' (\acc x -> p x || acc) False
```
ndmitchell: These definitions should be short-circuiting - so once you had any element which satisfies `p` you abort. You can't do that on top of `foldl` - you'll probably need a loop.

Foundation/Array/Chunked/Unboxed.hs
```diff
@@ -55,14 +55,20 @@ instance PrimType ty => IsList (ChunkedUArray ty) where
...
 instance PrimType ty => C.Collection (ChunkedUArray ty) where
     null = null
     length = length
     elem   = elem
     minimum = minimum
     maximum = maximum
-    all p = Data.List.all p . toList
-    any p = Data.List.any p . toList
+    all p = foldl' (\acc x -> p x && acc) True
+    any p = foldl' (\acc x -> p x || acc) False
```
ndmitchell: Ditto the comments above.

Foundation/Array/Chunked/Unboxed.hs
```diff
@@ -138,13 +144,26 @@ elem el (ChunkedUArray array) = loop 0
...
+
+foldl' :: PrimType ty => (a -> ty -> a) -> a -> ChunkedUArray ty -> a
+foldl' f initialAcc (ChunkedUArray cua) = A.foldl' (U.foldl' f) initialAcc cua
+
+foldr :: PrimType ty => (ty -> a -> a) -> a -> ChunkedUArray ty -> a
+foldr f initialAcc (ChunkedUArray cua) = A.foldr (flip $ U.foldr f) initialAcc cua
+
 minimum :: (Ord ty, PrimType ty) => C.NonEmpty (ChunkedUArray ty) -> ty
-minimum = Data.List.minimum . toList . C.getNonEmpty
+minimum cua = foldl' min (unsafeIndex cua' 0) cua'
```
ndmitchell: foldl1' here?

Foundation/Array/Bitmap.hs
ThreeFx: As far as I searched there is no `foldl1'`.

ndmitchell: It should definitely be added to foundation then. An oversight it isn't in base in my opinion!

ndmitchell: Just found it in Data.List right next to foldl'

Foundation/Array/Bitmap.hs
ThreeFx: Won't a short-circuiting `&&` or `||` terminate the fold prematurely?

ndmitchell: Hmm, it does seem to. Fair enough then - I guess it's worth adding an explicit test of this property somewhere.

ThreeFx:
The problem with `foldl1` is that it expects (or rather should expect) a `NonEmpty` foldable - something which doesn't exist in `Foundation.Collection.Foldable` but in `Foundation.Collection.Collection`. Or am I missing a branch in which it exists?

Foundation/Array/Bitmap.hs
```diff
@@ -443,5 +446,23 @@ foldl' f initialAcc vec = loop 0 initialAcc
         | i .==# len = acc
         | otherwise  = loop (i+1) (f acc (unsafeIndex vec i))
 
+all :: (Bool -> Bool) -> Bitmap -> Bool
+all p bm = loop 0
```
ndmitchell: Use `azero` instead of `0`?

Foundation/Array/Bitmap.hs
NicolasDP: agree, create #318

ThreeFx:
Okay, this is a bit bigger:

- There now exists a `Fold1able` typeclass (name may be changed) for folds on nonempty datastructures.
- Moved the `NonEmpty` type from `Collection.Collection` to new file `Collection.NonEmpty` to be able to write `fold1`s with correct type signature for `Block` etc.
- Wrote `foldl1`s etc. Mostly based on `fold` + a `splitAt` or `revSplitAt`

There now exists an orphan instance for `Collection c => Collection (NonEmpty c)`. Can this somehow be avoided?

Foundation/Array/Chunked/Unboxed.hs
ThreeFx: `foldl1'` for this data type is rather complex because the contained arrays may or may not be empty. I am pondering the best way to write this. Currently I just work around this by supplying the first element as accumulator and dropping it form the array.

Foundation/Array/Bitmap.hs
ThreeFx: The rest of the code (elem etc.) all use `0`. Is it possible that this is already optimized?

Foundation/Collection/Foldable.hs
```diff
@@ -43,6 +45,20 @@ class Foldable collection where
     foldr' :: (Element collection -> a -> a) -> a -> collection -> a
     foldr' f z0 xs = foldl f' id xs z0 where f' k x z = k $! f x z
 
+-- | Fold1's. Like folds, but they assume to operate on a NonEmpty collection.
+class Foldable f => Fold1able f where
```
NicolasDP: just extend `Foldable` with the `NonEmpty`.

Foundation/Collection/Foldable.hs
```diff
@@ -43,6 +45,20 @@ class Foldable collection where
...
+    -- | Left associative lazy fold.
+    foldl1  :: (Element f -> Element f -> Element f) -> NonEmpty f -> Element f
```
NicolasDP: we are actually considering removing the lazy `foldl` ( #319). Might be worth removing the lazy `foldl1` too. What do you think ?

Foundation/Collection/Collection.hs
```diff
@@ -31,18 +31,13 @@ module Foundation.Collection.Collection
...
 import qualified Foundation.Array.Unboxed as UV
 import qualified Foundation.Array.Boxed as BA
 import qualified Foundation.String.UTF8 as S
 
--- | NonEmpty property for any Collection
---
--- This can only be made, through the 'nonEmpty' smart contructor
-newtype NonEmpty a = NonEmpty { getNonEmpty :: a }
-    deriving (Show,Eq)
-
```
NicolasDP: I think you could keep it there and make `Foundation.Collection.Foldable` imports `Foundation.Collection.Collection` to get the `NonEmpty a` type

Foundation/Collection/Foldable.hs
ThreeFx: If `foldl` is removed then `foldl1` should definitely go too. But they have differing behaviour: `foldl (flip const) False [True, undefined, True] == True /= undefined = foldl' (flip const) False [True, undefined, True]`.

Foundation/Collection/Collection.hs
ThreeFx: Then I have circular dependencies because the boxed and unboxed array modules need the NonEmpty type for writing `fold1`s.

Foundation/Collection/Foldable.hs
ThreeFx: Some types aren't effectively `Fold1able`, like trees or the like because they have no canonical way of unconsing an element.

Foundation/Collection/Foldable.hs
NicolasDP: indeed. Still, might be worth removing it. I don't believe anyone would use the lazy version of `foldl` or `foldl1` properly anyway. There is a big warning on `base`'s `foldl`... some (including me) believe we should better remove it.

Foundation/Collection/Collection.hs
NicolasDP: indeed... damn...  😞

Foundation/Collection/Foldable.hs
ThreeFx: Alright, will do so then!

Foundation/Array/Bitmap.hs
vincenthz: no, it's not yet optimised; it's still going to allocate through an integer. let's not bother about this here as it's not that critical; I'll look either into fixing the issue at the source of the problem, or by maybe doing some kind of global replace of 0 to azero later

Foundation/Primitive/Block.hs
```diff
@@ -164,6 +167,14 @@ foldl' f initialAcc vec = loop 0 initialAcc
         | i .==# len = acc
         | otherwise  = loop (i+1) (f acc (unsafeIndex vec i))
 
+foldl1' :: PrimType ty => (ty -> ty -> ty) -> NonEmpty (Block ty) -> ty
+foldl1' f arr = let (initialAcc, rest) = splitAt 1 $ getNonEmpty arr
```
vincenthz: `splitAt` is not cheap in Block, because Block don't support slicing, so that mean a full copy - 1 of the block.
I think it would be simpler just to unroll the loop directly and only work on indexes

vincenthz:
Looking good overall, Thanks ! There's couple of niggle and performance optimisation that we can make later

PR #314, opened 2017-06-10
Fixed #289 (String.replace function)

### LINKED ISSUE #289: Add String.replace
It would be useful to have a `replace` function, with the semantics `replace "ab" "foo" "testabab" == "testfoofoo"`. Such a function exists in `text` and `extra`, both named `replace`. Writing such a function on top of foundation is possible (just use list, extra.replace, back again) - but not very performant. For extra performance, you may wish to optimise the common case when the from string is a single ASCII char.

adinapoli:
This PR adds support for the `String.replace` function, with the familiar type signature:

```
replace :: Needle -> Replacement -> Haystack -> New
```

It uses an approach similar to what [text does](https://github.com/bos/text/blob/c29d57cfd7f9b6236b9ec6f8c0e5bcaa98902d61/Data/Text.hs#L691): We first get all the `indices` where the replacement will happen (called `insertionPoints` here) and then we proceed to do the insertion.

Note the algorithm is still slow in a couple of places (but I expect performance to be decent enough to start with):

* We allocate an intermedia array inside `loop`, which we should scrap in favour of a byte-to-byte, zero-alloc equality function;
* Inside `indices` we right append; we should use something like a DList but here I couldn't due to circular dependencies error.

Let me know what you guys think.

### DIFF
--- a/Foundation/Array/Unboxed.hs
+++ b/Foundation/Array/Unboxed.hs
@@ -1031,21 +1031,23 @@ reverse a
 
 -- Finds where are the insertion points when we search for a `needle`
 -- within an `haystack`.
+-- Throws an error in case `needle` is empty.
 indices :: PrimType ty => UArray ty -> UArray ty -> [Offset ty]
-indices ned hy =
-  case haystackLen < needleLen of
-    True  -> []
-    False -> go (Offset 0) []
+indices needle hy
+  | needleLen <= 0 = error "Foundation.Array.Unboxed.indices: needle is empty."
+  | otherwise = case haystackLen < needleLen of
+                  True  -> []
+                  False -> go (Offset 0) []
   where
     !haystackLen = length hy
 
-    !needleLen = length ned
+    !needleLen = length needle
 
     go currentOffset ipoints
       | (currentOffset `offsetPlusE` needleLen) > (sizeAsOffset haystackLen) = ipoints
       | otherwise =
         let matcher = take needleLen . drop (offsetAsSize currentOffset) $ hy
-        in case matcher == ned of
+        in case matcher == needle of
              -- TODO: Move away from right-appending as it's gonna be slow.
              True  -> go (currentOffset `offsetPlusE` needleLen) (ipoints <> [currentOffset])
              False -> go (currentOffset + Offset 1) ipoints
--- a/Foundation/Array/Unboxed.hs
+++ b/Foundation/Array/Unboxed.hs
@@ -1055,11 +1055,14 @@ indices ned hy =
 -- the `haystack` string.
 replace :: PrimType ty => UArray ty -> UArray ty -> UArray ty -> UArray ty
 replace (needle :: UArray ty) replacement haystack = runST $ do
-    let insertionPoints = indices needle haystack
-    let !occs           = Prelude.length insertionPoints
-    let !newLen         = haystackLen - (multBy needleLen occs) + (multBy replacementLen occs)
-    ms <- new newLen
-    loop ms (Offset 0) (Offset 0) insertionPoints
+    case null needle of
+      True -> error "Foundation.Array.Unboxed.replace: empty needle"
+      False -> do
+        let insertionPoints = indices needle haystack
+        let !occs           = Prelude.length insertionPoints
+        let !newLen         = haystackLen - (multBy needleLen occs) + (multBy replacementLen occs)
+        ms <- new newLen
+        loop ms (Offset 0) (Offset 0) insertionPoints
   where
 
     multBy (CountOf x) y = CountOf (x * y)
--- a/Foundation/Array/Unboxed.hs
+++ b/Foundation/Array/Unboxed.hs
@@ -72,6 +72,7 @@ module Foundation.Array.Unboxed
     , break
     , breakElem
     , elem
+    , indices
     , intersperse
     , span
     , cons
@@ -82,6 +83,7 @@ module Foundation.Array.Unboxed
     , sortBy
     , filter
     , reverse
+    , replace
     , foldl
     , foldr
     , foldl'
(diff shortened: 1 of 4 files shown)


Foundation/String/UTF8.hs
```diff
@@ -820,6 +822,76 @@ reverse s@(String ba) = runST $ do
                 _  -> return () -- impossible
             loop ms (si `offsetPlusE` nb) d
 
+-- Finds how many times we need to perform the replacement, and where are
+-- the insertion points.
+indices :: String -> String -> (Int, [Offset8])
```
ndmitchell: The type of this is a big weird. I'd have expected indices, and not the number of them.

Foundation/String/UTF8.hs
```diff
@@ -820,6 +822,76 @@ reverse s@(String ba) = runST $ do
...
+    !needleLen = size needle
+
+    go :: Offset8 -> (Int, [Offset8]) -> (Int, [Offset8])
+    go currentOffset (!occs, !ipoints)
+      | (currentOffset `offsetPlusE` needleLen) > (sizeAsOffset haystackLen) = (occs, ipoints)
+      | otherwise =
+        let matcher = Vec.take needleLen . Vec.drop (offsetAsSize currentOffset) $ hy
+        in case matcher == ned of
+             -- TODO: Move away from right-appending as it's gonna be slow.
+             True  -> go (currentOffset `offsetPlusE` needleLen) (occs + 1, ipoints <> [currentOffset])
```
ndmitchell: Not returning the Int probably solves this.

Foundation/String/UTF8.hs
```diff
@@ -820,6 +822,76 @@ reverse s@(String ba) = runST $ do
                 _  -> return () -- impossible
             loop ms (si `offsetPlusE` nb) d
 
+-- Finds how many times we need to perform the replacement, and where are
+-- the insertion points.
+indices :: String -> String -> (Int, [Offset8])
+indices needle@(String ned) haystack@(String hy) =
```
ndmitchell: Suggest you generalise this to UArray and implement the String one via newtype wrapping on that.

Foundation/String/UTF8.hs
```diff
@@ -820,6 +822,76 @@ reverse s@(String ba) = runST $ do
...
+  case haystackLen < needleLen of
+    True  -> (0, [])
```
ndmitchell: needleLen of 0 will result in an infinite loop - it should raise an error saying 0 length is disallowed.

tests/Test/Foundation/String.hs
```diff
@@ -52,6 +52,30 @@ testStringCases =
...
+            replace "aa" "bb" "aabb" @?= "bbbb"
+        , testCase "replace 'aa' 'bb' 'aaccaadd' == 'bbccbbdd'" $ do
+            replace "aa" "bb" "aaccaadd" @?= "bbccbbdd"
+        , testCase "replace 'aa' 'LongLong' 'aaccaadd' == 'LongLongccLongLongdd'" $ do
+            replace "aa" "LongLong" "aaccaadd" @?= "LongLongccLongLongdd"
+        , testCase "replace 'aa' 'bb' 'aabbccabbccEEaaaaabb' == 'bbbbccabbccEEbbbbabb'" $ do
+            replace "aa" "bb" "aabbccabbccEEaaaaabb" @?= "bbbbccabbccEEbbbbabb"
+        , testCase "replace '😀' '😺' '😀 🙂 😉 😌' == '😺 🙂 😉 😌'" $ do
+            replace "😀" "😺" "😀 🙂 😉 😌" @?= "😺 🙂 😉 😌"
+                          ]
```
ndmitchell: Add a test of empty needle.

adinapoli:
Thanks for the great feedback, @ndmitchell !

It all sounds sensible.

> Add a test of empty needle.

I was discussing this with @vincenthz and we were unsure what's the best semantic should be here. From one side, to stick to the principle of _least surprise_, we should do what `text` is doing, which is throwing an user error on an empty needle (as you suggested):

```
ghci> T.replace "" "foo" "babacar"
"*** Exception: Data.Text.replace: empty input
CallStack (from HasCallStack):
  error, called at ./Data/Text.hs:1756:18 in text-1.2.2.1-5QpmrLQApEZ4Ly9nMHWY0s:Data.Text
ghci>
```

On the other hand, Vincent was suggesting that _maybe_ we could make the needle argument a `NonEmpty`. The advantages is that, of course, we keep the function total, but at the cost of more API overhead, which might not be that great.

Do you guys have any thought?

ndmitchell:
I guess foundation makes partiality explicit, which usually tends to be a good thing, so non-empty seems reasonable? If I write `replace "x" "" ""` does that work with NonEmpty by default? Or would I need to explicitly NonEmpty there?

adinapoli:
Last time I looked at that module (but I'm not 100% sure) the only way to construct a `NonEmpty` is via the smart constructor, which can return `Nothing` if the `String` is empty. So the annoying bit would be for the user to pattern match on this extra layer of coating.

ndmitchell:
It would be doubly painful to have to fromJust it as well... There there be a partialNonEmpty helper?

adinapoli:
@ndmitchell I eventually went with the path of least resistance and make `replace` partial. I think that for the sake of this PR it should be enough, and we can change our mind later on in case we want to go via the `NonEmpty` route.

What do you guys think?

Foundation/Array/Unboxed.hs
```diff
@@ -1027,6 +1029,74 @@ reverse a
             | otherwise = primMbaWrite ma i (primAddrIndex ba (sizeAsOffset (endI - i))) >> loop (i+Offset 1)
 {-# SPECIALIZE [3] reverse :: UArray Word8 -> UArray Word8 #-}
 
+-- Finds where are the insertion points when we search for a `needle`
+-- within an `haystack`.
+indices :: PrimType ty => UArray ty -> UArray ty -> [Offset ty]
+indices ned hy =
+  case haystackLen < needleLen of
```
ndmitchell: I think indicies should be partial on empty arrays, otherwise it loops forever. That partiality should be explained in the doc comment.

adinapoli: Ah damn, well spotted!

adinapoli:
@ndmitchell Ok, we should be in business!

I ended up removing the funky emojis and I have replaced them with more common UTF8 runes, as they were breaking the Windows & 7.8 builds.

Foundation/Array/Unboxed.hs
```diff
@@ -1027,6 +1029,76 @@ reverse a
...
+-- Finds where are the insertion points when we search for a `needle`
+-- within an `haystack`.
+-- Throws an error in case `needle` is empty.
+indices :: PrimType ty => UArray ty -> UArray ty -> [Offset ty]
+indices needle hy
+  | needleLen <= 0 = error "Foundation.Array.Unboxed.indices: needle is empty."
+  | otherwise = case haystackLen < needleLen of
+                  True  -> []
+                  False -> go (Offset 0) []
+  where
```
ndmitchell: Style only, but I'd lift the case into the guard here.

Foundation/Array/Unboxed.hs
```diff
@@ -1027,6 +1029,76 @@ reverse a
...
+             True  -> go (currentOffset `offsetPlusE` needleLen) (ipoints <> [currentOffset])
+             False -> go (currentOffset + Offset 1) ipoints
+
+-- | Replace all the occurrencies of `needle` with `replacement` in
+-- the `haystack` string.
+replace :: PrimType ty => UArray ty -> UArray ty -> UArray ty -> UArray ty
+replace (needle :: UArray ty) replacement haystack = runST $ do
+    case null needle of
+      True -> error "Foundation.Array.Unboxed.replace: empty needle"
+      False -> do
```
ndmitchell: Should we optimise the case where there are no indices?

ndmitchell: Thinking about it - we definitely should (happy to merge now and ticket that later though).

adinapoli:
Hey guys, my preference would be merging now and amend later. I have
already left the hackathon (boohoo!) So merging it will give me the
illusion it was a productive one :D

vincenthz:
yes I'm fine with this; I think it looks sensible already, and we can fix details later.
also just for the record, I would prefer not to error in the case of the empty needle, but happy to go with what's there.

vincenthz:
any chance you can rebase it ? no worries if you can't, just wondering

adinapoli:
At the airport currently, if I'm not breaking any golden rule and I can go
by without rebasing, definitely go ahead and press the trigger ;)

vincenthz:
it's because it doesn't merge anymore that I'm asking that. that's fine I'll rebase it manually myself

adinapoli:
Ah drat I realised later that might have been the case! I'll try to rebase
later in the day, unless somebody will beat me on that ;)

vincenthz:
Thanks @adinapoli !

PR #323, opened 2017-06-10
Remove `foldl` from Foundation

ThreeFx:
Slightly adjusted the documentation of `Foldable`. There were 2 or 3 calls to `foldl` in the code - a slight glance did not reveal any anomalies.


NicolasDP:
Thanks, great stuff.

Issue #303, opened 2017-06-09
String builder


vincenthz:
having the ability to build a string in an efficient way

PR #324, opened 2017-06-10
Add really simple String Builder

vincenthz:
provide api for #303

### DIFF
--- a/Foundation/String/Builder.hs
+++ b/Foundation/String/Builder.hs
@@ -1,0 +1,57 @@
+-- |
+-- Module      : Foundation.String.Builder
+-- License     : BSD-style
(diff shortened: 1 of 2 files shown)


ndmitchell:
I'd have been tempted to introduce a vector builder, then have string builder just wrap it with a newtype layer. If you did that, emit might have been called emitElem rather than emitChar - which might provide a more general name. You might also think about making it `Builder a` where `a` is probably `String`?

vincenthz:
I agree @ndmitchell . the only reason I could imagine where it could make a difference to have, is to be able to cheat on the UTF8 encoding side of things, although I suppose we could just have an emitUtf8 in array of word8 and short circuit the problem.

PR #325, opened 2017-06-10
Add touch capability for Block

vincenthz:


Foundation/Primitive/Block/Base.hs
```diff
@@ -324,3 +326,10 @@ unsafeRead (MutableBlock mba) i = primMbaRead mba i
 unsafeWrite :: (PrimMonad prim, PrimType ty) => MutableBlock ty (PrimState prim) -> Offset ty -> ty -> prim ()
 unsafeWrite (MutableBlock mba) i v = primMbaWrite mba i v
 {-# INLINE unsafeWrite #-}
+
+touch :: PrimMonad prim => Block ty -> prim ()
+touch (Block ba) = primitive $ \s -> case touch# ba s of { s2 -> (# s2, () #) }
+
+mutableTouch :: PrimMonad prim => MutableBlock ty (PrimState prim) -> prim ()
+mutableTouch (MutableBlock mba) = primitive $ \s -> case touch# mba s of { s2 -> (# s2, () #) }
```
NicolasDP: I think you need to limit your functions to `RealWord#` as the signature of `touch#` is already specialised

```haskell
touch# :: o -> State# RealWorld -> State# RealWorld
```

Issue #316, opened 2017-06-10
Use Collection.any instead of reimplementing or


galderz:
Foundation/Check/Main.hs implements its own `or` function.

Make it use `Collection.any` instead.

NicolasDP:
one could implement `or` in `Foundation/Collection/Collection.hs`:

```haskell
or :: (Collection col, Element col ~ Bool) => col -> Bool
or = any ((==) True)
```

NicolasDP:
It would be great to add `and` too.

```haskell
and :: (Collection col, Element col ~ Bool) => col -> Bool
```

PR #330, opened 2017-06-11
And 'and' and 'or' functions to the Collection module

lortabac:
Fixes https://github.com/haskell-foundation/foundation/issues/316

### DIFF
--- a/Foundation/Collection/Collection.hs
+++ b/Foundation/Collection/Collection.hs
@@ -26,9 +26,11 @@ module Foundation.Collection.Collection
     , nonEmpty
     , nonEmpty_
     , nonEmptyFmap
+    , and
+    , or
     ) where
 
-import           Foundation.Internal.Base
+import           Foundation.Internal.Base hiding (and)
 import           Foundation.Primitive.Types.OffsetSize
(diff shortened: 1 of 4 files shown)


Foundation/Collection/Collection.hs
```diff
@@ -157,3 +159,11 @@ instance Collection c => Collection (NonEmpty c) where
...
     all p = all p . getNonEmpty
     any p = any p . getNonEmpty
+
+-- | Return True if all the elements in the collection are True
+and :: (Collection col, Element col ~ Bool) => col -> Bool
+and = all (== True)
+
+-- | Return True if at least one element in the collection is True
+or :: (Collection col, Element col ~ Bool) => col -> Bool
+or = any (== True)
```
NicolasDP: You might want to override the hlint error here below the [line 82 of of .hlint.yaml](https://github.com/haskell-foundation/foundation/blob/master/.hlint.yaml#L82)

vincenthz:
merged. thanks @lortabac !

CaptJakk:
It looks like this was closed rather than merged as the comment suggests. Should it be reopened?

Issue #331, opened 2017-06-11
Lines should either ignore '\r' or document it and have linesWithR


ndmitchell:
Since foundation does binary IO by default (a sensible choice), often `lines` will do the wrong thing unless you `filter (/= '\r')` first. Perhaps provide `linesWithR` or something that does normal lines plus filters out `\r`?

vincenthz:
using filter after the fact is probably much less efficient. I would definitely prefer a lines function that cut '\r\n' or/and '\n', although as a workaround for the non existant function we could lines + revDropWhile (== '\r'), (which obviously doesn't exist either, since apparently I do that :))

ndmitchell:
Thinking further, people rarely (if ever) want lines which leaves trailing `\r` at the end of lines, so suggest doing this in the standard `lines` function - otherwise the standard Windows compat bug will be changing all instances of `lines` to `linesWithCR`.

vincenthz:
I gave this more thoughts too and this is not straightforward:

* prelude lines only split on '\n' (i.e. unix mode)
* the textual handle mode for windows is putting everything in unix mode (at a cost of rewriting).

Since everything in foundation is going to binary that open a can of worm related to this:

1. provide a binary translation conduit that is only active on windows otherwise would just let things through on unix
2. make conduit lines system dependant (split on LF only on unix, and CRLF on windows)
3. make it a sourceHandleTextual derivative of sourceHandle which would do all possible rewriting (UTF16 -> UTF8, CRLF -> LF, ...) in one place. This could work either in autodetect mode (which UTF, which linefeed mode) or this could also expect specific mode (and error if not)

1. easy to forget to put the windows only rewriting when working on unix system, cost of rewriting "twice" when doing lines after that
2. more efficient, but if you process your lines yourself from multi lines (e.g. a parser), you need to keep this implementation detail in mind
3. less efficient on windows only, but most flexible and reduce the number of user level impact

vincenthz:
Also I'm tempted to expose String's `lines` with a non-optional parameter for the mode, e.g.:

```haskell
data LineMode = CRLF | LF | AutoDetect | SystemDependent
lines :: LineMode -> String -> [String]
```

ndmitchell:
A lot of developers test only on one platform. I think it's a reasonable goal that whatever we do should work for as many developers as possible, without requiring them to take a "make it work on Windows" special case patch.

[...]

PR #333, opened 2017-06-11
Add Conduit.sourceList

ndmitchell:
As @vincenthz suggested I use in #331 but didn't really exist.

### DIFF
--- a/Foundation/Conduit.hs
+++ b/Foundation/Conduit.hs
@@ -5,6 +5,7 @@ module Foundation.Conduit
     , await
     , awaitForever
     , yield
+    , yields
     , yieldOr
     , leftover
     , runConduit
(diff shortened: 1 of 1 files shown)


NicolasDP:
Vincent does that often.

NicolasDP:
This could be generalised to `sourceCollection` constrained to `Mappable`.

ndmitchell:
I'd ideally use `mapM_` to do that, but it turns out `mapM_` in foundation is all kinds of wrong (see #334). Perhaps include this as is, and generalising to Mappable in future?

At the same time, you probably want to stick with sinkList (it's not generalisable in the same easy ways), so maybe for symmetry it should always be sourceList?

ndmitchell:
I added `yields` which is generalised to any `Foldable`.

PR #338, opened 2017-06-11
Enable `-Werror` on GHC 8.0 by default

ThreeFx:
4 outstanding warnings fixed or surpressed:
 - `inline-rule-shadowing` was removed using `GHC_OPTIONS` pragma
 - `orphan` as well (needed in that particular case) 
 - `unused-import` was fixed

### DIFF
--- a/Foundation/Conduit/Internal.hs
+++ b/Foundation/Conduit/Internal.hs
@@ -11,6 +11,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# OPTIONS_GHC -Wno-inline-rule-shadowing #-}
 module Foundation.Conduit.Internal
     ( Pipe(..)
     , Conduit(..)
--- a/Foundation/Conduit/Internal.hs
+++ b/Foundation/Conduit/Internal.hs
@@ -11,7 +11,11 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE OverloadedStrings #-}
+
+#if __GLASGOW_HASKELL__ >= 800
 {-# OPTIONS_GHC -Wno-inline-rule-shadowing #-}
+#endif
+
 module Foundation.Conduit.Internal
     ( Pipe(..)
     , Conduit(..)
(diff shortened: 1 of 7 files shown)


ThreeFx:
Apparently, the windows version has an unused import as well. I'll see if I can do something about that.

Foundation/String/ASCII.hs
```diff
@@ -17,6 +17,7 @@
 {-# LANGUAGE TypeFamilies               #-}
 {-# LANGUAGE UnboxedTuples              #-}
 {-# LANGUAGE FlexibleContexts           #-}
+{-# OPTIONS_GHC -Wno-inline-rule-shadowing #-}
```
NicolasDP: old compilers doesn't like that one...

Foundation/Conduit/Internal.hs
```diff
@@ -11,6 +11,7 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# OPTIONS_GHC -Wno-inline-rule-shadowing #-}
```
NicolasDP: old compilers doesn't like that one...

ThreeFx:
Regarding the `break` rule: GHC is at liberty to rewrite that statement as a case statement and I see no way of blocking that from happening (see `litEq` in `PrelRules.lhs` of the GHC documentation. I don't know what's best - I suppose we could ignore the warning and add a custom HLint hint as `use breakElem`.

No on to the other one...

ThreeFx:
Same thing for `maybe`. It's from `Data.Maybe` and since it's a small function, GHC may inline `maybe` or may not - I don't have enough experience in deciding when the compiler does what.
Maybe someone can either look at this himself or point me in the right direction?

ThreeFx:
I don't particularily like the macros as a way to reduce these minimal errors on OS X and Windows. What do you suggest instead?

NicolasDP:
You know what, don't try to fix this issues and only enable the `-Werror`, you will need update the `travis-ci` file to pass it only when building with ghc8.0. This will fix your error.

vincenthz:
I originally thought that it would be a terrible idea to redefine maybe, but now I think the consequence it's not too bad; namely if you use prelude's maybe you get no optimisation anymore for sure instead of cross-your-fingers-optimisations.

Also I think it probably make sense to expose `awaitMaybe` directly too.

adding @snoyberg just in case there's a reason it's just a stupid idea to redefine maybe ourselves ..

NicolasDP:
what about compatibility and interoperability with `base`'s `Maybe` ?

ThreeFx:
I think @vincenthz doesn't want to redefine the datatype, but rather just the `maybe` function to have control of inlining and rewrite rules.

NicolasDP:
The PR has staled, I will close it, we can pick this up later.

PR #339, opened 2017-06-11
Prime Types faster arbitrary

NicolasDP:
This will make `:check-foundation` test suite run 2 times faster.

For the `Floating` points and the `Double` random number generator I am masking the results of the `Word32` and `Word64` random generator to a _sensible_ scope.

This generates more precise number generator but now I get the following error:

```
Seed: 3243238201641105201

foundation
  String
    reading
       ✗ Double failed after 1 test
         ✗ Prelude.read failed after 3 tests:
           use param: --seed 3243238201641105201
parameter 1 : 4.6114233520297733e18
Property `cond1 && cond2' failed where:
   cond1 = Property `a -1.0e-8 < 4.6114233520297733e18 - 4.611423352029774e18 (== -512.0 ) b' failed where:

               a = -1.0e-8 :: Double
                    ^^^^^^^ ^^^^^^^^
               b = -512.0 :: Double
                    ^^^^^^^ ^^^^^^^
   cond2 = Succeed

Failed 1 out of 1

Test suite failure for package foundation-0.0.10
    check-foundation:  exited with: ExitFailure 1
```

1. The new random generator algorithm generates `4.6114233520297733e18`
2. `Prelude.show` gave `"4.6114233520297733e18"`
3. `Foundation.readDouble` gave `4.6114233520297733e18`
4. `Prelude.read` gave `4.611423352029774e18`

i.e.:

```
Foundation: 4.6114233520297733e18
Prelude:    4.6114233520297740e18
                             ^
```

This is obviously... intriguing... Did I change something totally wrong ?

@vincenthz what do you think ?

### DIFF
--- a/Foundation/Random.hs
+++ b/Foundation/Random.hs
@@ -44,10 +44,16 @@ import qualified Foreign.Marshal.Alloc (alloca)
 class (Functor m, Applicative m, Monad m) => MonadRandom m where
     getRandomBytes :: CountOf Word8 -> m (UArray Word8)
     getRandomWord64 :: m Word64
+    getRandomF32 :: m Float
+    getRandomF64 :: m Double
 
 instance MonadRandom IO where
     getRandomBytes  = getEntropy
     getRandomWord64 = flip A.index 0 . A.unsafeRecast
+                  <$> getRandomBytes (A.primSizeInBytes (Proxy :: Proxy Word64))
+    getRandomF32 = flip A.index 0 . A.unsafeRecast
+                  <$> getRandomBytes (A.primSizeInBytes (Proxy :: Proxy Word64))
+    getRandomF64 = flip A.index 0 . A.unsafeRecast
                   <$> getRandomBytes (A.primSizeInBytes (Proxy :: Proxy Word64))
 
 -- | A Deterministic Random Generator (DRG) class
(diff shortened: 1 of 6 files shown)


Foundation/Random.hs
```diff
@@ -37,13 +38,23 @@ import qualified Foundation.Array.Unboxed as A
 import qualified Foundation.Array.Unboxed.Mutable as A
 import           GHC.ST
 import qualified Prelude
+import qualified Foreign.Marshal.Alloc (alloca)
 
 -- | A monad constraint that allows to generate random bytes
 class (Functor m, Applicative m, Monad m) => MonadRandom m where
     getRandomBytes :: CountOf Word8 -> m (UArray Word8)
+    getRandomWord64 :: m Word64
+    getRandomF32 :: m Float
```
vincenthz: I'm tempted to just make this just a downsize of `Double` (calling the doubleToFloat function that is available under some name I can't remember from ghc prim) instead of making it part of the class

ndmitchell:
Modern code only uses float32 in vectors so foundation making it second class makes sense.

vincenthz:
that does looks like a prelude bug indeed !

vincenthz:
this is a round trip printing+parsing using printf&strtod in C:

```
first print = 4.6114233520297733e+18
print after parse = 4.6114233520297733e+18
```

we look more precise in this specific instance ..

vincenthz:
rebased and fix the test to scale the differential limit according to the number tested.

PR #349, opened 2017-06-14
Fix #312: Refactor Builder to support possibility of failure.

### LINKED ISSUE #312: `fromBytes` throws exceptions instead of returning the error inside the `Maybe ValidationFailure`
The following program throws a `ByteOutOfBound` exception instead of providing the error in the `Maybe ValidationError` in the triple returned by `fromBytes`.

```haskell
#!/usr/bin/env stack
-- stack script --resolver lts-8.17 --package foundation
{-# LANGUAGE NoImplicitPrelude #-}
{-# LANGUAGE OverloadedStrings #-}
module Main where

import Control.Monad
import Foundation
import Foundation.String
import Foundation.String.ASCII
import Foreign.C.Types

printString :: Encoding -> UArray Word8 -> IO ()
printString enc bytes = do
  -- fromBytes :: Encoding -> UArray Word8 -> (String, Maybe ValidationFailure, UArray Word8)
  let (string, minvalid, rest) = fromBytes enc bytes
  putStrLn string
  case minvalid of
    Nothing -> putStrLn "Valid!"
    (Just vf) -> putStrLn "Invalid!"
  when (null rest) $ do
    putStrLn "rest is empty"
  unless (null rest) $ do
    putStrLn "rest is not empty"

main = printString ASCII7 (fromList [0xff])
```

```shell
strings: ByteOutOfBound 255
```

jship:
Started during ZuriHac 2017 with @Tiltman.  Fixes #312.

### DIFF
--- a/Foundation/String/UTF8.hs
+++ b/Foundation/String/UTF8.hs
@@ -67,6 +67,7 @@ module Foundation.String.UTF8
     , replace
     , builderAppend
     , builderBuild
+    , builderBuild_
     , readInteger
     , readIntegral
     , readNatural
@@ -126,6 +127,7 @@ import qualified Foundation.String.Encoding.ISO_8859_1 as Encoder
 data ValidationFailure = InvalidHeader
                        | InvalidContinuation
                        | MissingByte
+                       | BuildingFailure
                        deriving (Show,Eq,Typeable)
 
 instance Exception ValidationFailure
@@ -279,7 +281,7 @@ nextWithIndexer getter off =
 
 writeWithBuilder :: (PrimMonad st, Monad st)
                  => Char
-                 -> Builder (UArray Word8) (MVec.MUArray Word8) Word8 st ()
+                 -> Builder (UArray Word8) (MVec.MUArray Word8) Word8 st err ()
 writeWithBuilder c
     | bool# (ltWord# x 0x80##   ) = encode1
     | bool# (ltWord# x 0x800##  ) = encode2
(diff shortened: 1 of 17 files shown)


Foundation/String/UTF8.hs
```diff
@@ -872,17 +874,19 @@ data Encoding
...
                  -> UArray Word8
                  -> (String, Maybe ValidationFailure, UArray Word8)
 fromEncoderBytes enc bytes =
-    ( String $ runST $ Encoder.convertFromTo enc EncoderUTF8 (Vec.recast bytes)
-    , Nothing
-    , mempty
-    )
+    case runST $ Encoder.convertFromTo enc EncoderUTF8 (Vec.recast bytes) of
+        -- TODO: Don't swallow up specific error (second element of pair)
+        -- TODO: Confused why all this recasting is necessary. I "typed hole"-ed my way to get this function to compile.  Feels like there should be a cleaner method.
```
vincenthz: we discussed this, and the original `recast` was a bad idea in the first place. so we'll sort this out with a overall of the decoder that I outlined in #353

In this case instead of recasting the array, you should downcast the `Offset` back to `Word8` unit, so that you can splice the original array in 2 more efficiently.

vincenthz:
looking good guys ! thanks @jship @Tiltman !

I've raised #354 #353 #352 #351 that got uncovered while looking at the issues you raised (specially the recasting), well done !

PR #350, opened 2017-06-14
String optimisations

vincenthz:
* Add takeWhile and dropWhile
* make uncons & unsnoc more efficient
* Add primitive to process from the end
* Optimise array filter
* Massive speedup in String filter & validation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 .stack-work
 exts
 dist
+_site
 report.html
 cabal.config
 .idea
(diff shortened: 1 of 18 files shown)


vincenthz:
@ndmitchell this should bring a massive speedup in your weeder experiment

ndmitchell:
Cool, so concretely for my use case:

* Use `lines` from conduit.
* Use `uncons` to strip away the `\r`.
* Validate goes much faster, so hopefully that's not a bottleneck.

Will try it later and report back.

Issue #358, opened 2017-06-16
Failure in ghci on Windows


ndmitchell:
Currently failing like #326, but this time with `foundation_time_clock_gettime`.

RyanGlScott:
I've opened #360 for this.

PR #359, opened 2017-06-17
Improve the appveyor script

ndmitchell:
Does 3 things:

* Removes entirely redundant cabal download and extract (it wasn't ever needed).
* Simplifies the test stuff.
* Add a test that GHC isn't bust on Windows. This currently fails due to #358, which is a good thing.

I tried to run the benchmarks on the Appveyor server, but it takes too long. Perhaps if anyone were reinventing a benchmark running system they'd give the ability to set a target execution time, so you can run benchmarks quickly.

(You probably want to squash these commits - a lot of it was experimenting on Appveyor)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -15,6 +15,5 @@ environment:
 
 test_script:
 - stack setup > nul
-- echo "" | stack --no-terminal build
-- stack test
-- stack exec -- ghci -e "Fondation.Tuple2 1 1"
+- echo "" | stack --no-terminal build --test --bench
+- stack exec -- ghci -e "Foundation.Tuple2 1 1"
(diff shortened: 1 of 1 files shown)


ndmitchell:
Currently fails with:

```
ghc.EXE: unable to load package `foundation-0.0.11'
ghc.EXE: C:\project\.stack-work\install\e56c21fa\lib\i386-windows-ghc-8.0.2\foundation-0.0.11-OhecGJQgBx9kCNSeKPAcs\HSfoundation-0.0.11-OhecGJQgBx9kCNSeKPAcs.o: unknown symbol `_foundation_time_clock_gettime'
```

vincenthz:
Thanks ! I'm merging to master since master is effectively broken anyway

Issue #326, opened 2017-06-11
Failure in ghci on Windows


ndmitchell:
I modified weeder to use foundation strings. I loaded it in ghci, but running gives:

```
ghc.exe: C:\Users\Neil\AppData\Roaming\cabal\x86_64-windows-ghc-8.0.2\foundation-0.0.10-2WPaw9Yl85PEz44LOCUDJh\HSfoundation-0.0.10-2WPaw9Yl85PEz44LOCUDJh.o: unknown symbol `clock_gettime'
```

I can reproduce with `runhaskell Test.hs` where `Test.hs` is defined as:

```
import Foundation as F
main = F.putStrLn (F.fromString "Hello")
```

I suggest adding that to the test suite.

ndmitchell:
See https://stackoverflow.com/questions/5404277/porting-clock-gettime-to-windows for how to avoid calling clock_time.

RyanGlScott:
FWIW, this is an occurrence of a known GHC bug, [Trac #13120](https://ghc.haskell.org/trac/ghc/ticket/13210), which has been fixed in GHC 8.2.1. A workaround is to link against `pthread` explicitly:

```
$ runghc Bug.hs
ghc.exe: C:\Users\RyanGlScott\AppData\Roaming\cabal\x86_64-windows-ghc-8.0.2\foundation-0.0.10-2WPaw9Yl85PEz44LOCUDJh\HSfoundation-0.0.10-2WPaw9Yl85PEz44LOCUDJh.o: unknown symbol `clock_gettime'

Bug.hs: Bug.hs: unable to load package `foundation-0.0.10'

$ runghc -lpthread Bug.hs
Hello
```

NicolasDP:
@RyanGlScott , thanks for the heads up. Didn't it was a GHC bug.

vincenthz:
I wonder if it's a cabal only problem. I've extensively tested on windows 7 & 10 before pushing this to master (albeit all with stack)

vincenthz:
ok I managed to reproduce using runghc, sadly I have no idea how to properly switch to a compat definition when needed; it works with `clock_gettime` for an unknown reason when compiling a program with stack, which I expect something about mingw that get linked on or some include, no idea ..

RyanGlScott:
Right, it's specifically an issue with the runtime linker, which is typically only invoked from GHCi or `runghc` (and not from compiled code, i.e., `cabal build`/`stack build`).

[...]

PR #360, opened 2017-06-17
Fix #358

### LINKED ISSUE #358: Failure in ghci on Windows

RyanGlScott:
Previously, `foundation_time_clock_gettime` wasn't actually being compiled since the C file it was defined in `foundation_time.c` was checking for the presence of the `FOUNDATION_SYSTEM_API_NO_CLOCK` macro. But this macro was defined in a Haskell file, `Time.hsc`! As a result, the macro was never detected in `foundation_time.c`.

Due to infelicities of how the runtime linker works, it would attempt to eagerly load the `foundation_time_clock_gettime` symbol on programs that didn't use it, causing strange linker errors like the ones observed in #358 and https://github.com/haskell-foundation/foundation/issues/326#issuecomment-309219955.

This shuffles around the `#define`s so that they're all in `foundation_system.h` where they belong. This also adds a test which uses Template Haskell to confirm that the runtime linker really does the right thing here.

### DIFF
--- a/Foundation/System/Bindings/Time.hsc
+++ b/Foundation/System/Bindings/Time.hsc
@@ -11,6 +11,7 @@ import Foreign.C.Types
 
 #include <time.h>
 #include <sys/time.h>
+#include "foundation_system.h"
 
 type CClockId = CInt
 data CTimeSpec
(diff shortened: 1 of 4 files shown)


ndmitchell:
Note that I added a quite different test for the same issue in https://github.com/haskell-foundation/foundation/pull/359. Having both tests seems even better.

RyanGlScott:
The AppVeyor build is failing for reasons I don't understand, but at the very least, I can confirm that this patch works because the tests are _linking_ correctly, which is the nub of this PR.

RyanGlScott:
By the way, the "infelicities" I referred to in https://github.com/haskell-foundation/foundation/pull/360#issue-236669545 have been reported upstream at https://ghc.haskell.org/trac/ghc/ticket/13845. Until that is fixed, this patch is necessary.

ndmitchell:
I fixed the Appveyor in #362.

vincenthz:
thanks @RyanGlScott; I will probably move the template haskell test to its own place later to prevent having template haskell as requirement for all tests.

PR #361, opened 2017-06-17
Add Base64 (fix #138)

### LINKED ISSUE #138: Add Base64

jship:
Add support for Base64 - see issue #138.  Modeled off of the Base64 code in the `memory` package and converted to work with `UArray`/`String`.

Property tests available for length/content as well as a handful of manual unit tests.

### DIFF
--- a/Foundation/Array/Unboxed.hs
+++ b/Foundation/Array/Unboxed.hs
@@ -96,8 +96,10 @@ module Foundation.Array.Unboxed
     , builderBuild
     , builderBuild_
     , toHexadecimal
+    , toBase64Internal
     ) where
 
+import           Control.Monad (when)
 import           GHC.Prim
 import           GHC.Types
 import           GHC.Word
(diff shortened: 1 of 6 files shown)


Foundation/Array/Unboxed.hs
```diff
@@ -1252,3 +1254,71 @@ toHexadecimal ba
...
+
+                loop (sIdx `offsetPlusE` CountOf 3) (dIdx `offsetPlusE` CountOf 4)
+
+outputLengthBase64 :: Bool -> CountOf Word8 -> CountOf Word8
+outputLengthBase64 padding (CountOf inputLenInt) = outputLength
+  where
+    outputLength = if padding then CountOf lenWithPadding else CountOf (lenWithPadding - numPadChars)
+
+    lenWithPadding :: Int
+    lenWithPadding = 4 * roundUp (Prelude.fromIntegral inputLenInt / 3.0 :: Double)
```
vincenthz: I think this would probably works without a hitch, but I would stay away from floating point operation if not necessary and use `divMod` which is going to give you the divisor and the numPadChars in 1 operation.

Foundation/Array/Unboxed.hs
```diff
@@ -1252,3 +1254,71 @@ toHexadecimal ba
...
+    !endOfs = Offset 0 `offsetPlusE` len
+
+    go :: MUArray Word8 s -> (Offset Word8 -> Word8) -> ST s ()
+    go !ma !getAt = loop 0 0
+      where
+        eqChar = 0x3d :: Word8
+
+        loop !sIdx !dIdx
+            | sIdx >= endOfs = return ()
+            | otherwise = do
```
vincenthz: I think here, I would put the case where you going to write padding (1 & 2 bytes left) too, and that would leave the `otherwise` (i.e. at least 3 bytes) case clearer and the only case that would call `loop` again.

```
loop
    | sIdx == endOfs = return ()
    | sIdx + 3 > endOfs = ... -- no 3 bytes, padding cases
    | otherwise          = ... -- normal cases, + looping
```

vincenthz:
thanks. well done, apart from minor comment it's fine !

Foundation/Array/Unboxed.hs
jship: Good call - this cleans up the ifs nicely.   Available in #363.

Foundation/Array/Unboxed.hs
jship: `divMod` is a much better approach than using floating point.  Available in #363.  Thanks!

PR #362, opened 2017-06-18
Improve the appveyor script, avoid an error when it does work

ndmitchell:
Turns out whether `Foundation.Foo` works or not is variable. Fortunately just loading foundation via `-package` reproduces the bug, so I'm reducing it to that.


ndmitchell:
The appveyor here will probably fail until #360 is fixed, but then should work.

PR #363, opened 2017-06-20
Base64 updates per PR #361 review

jship:
Base64 updates/refactoring per PR #361's review.


vincenthz:
it's not quite what I was thinking, but that's fine too :) thanks !

Issue #293, opened 2017-06-09
Change Difference type in CountOf to return a Maybe


vincenthz:
we should only have CountOf that are positive, and the only way to create negative count of would be using the `(-)` operation. By having `(-)` returning a `Maybe CountOf`, we prevent most negative count bug.

also having this we can replace the code that uses `(-)` on CountOf which is usually : 

```
if n >= m
    then ..ok,we have element to remove.. now the size is n - m..
    else ..do_something_not_enough_data..
```

into something that is safer by construction:

```
case n - m of
    Nothing -> ..do_something_not_enough_data
    Just newSize -> ...remove_something_and_then_newSize...
```

madidier:
I'm taking this ticket

CaptJakk:
Can this be closed?

PR #376, opened 2017-07-04
Difference of count returns maybe

ciderale:
I completed my work on #293 that I started at ZuriHac 2017 and I'm looking forward to your feedback.

The difference type for `CountOf` has been changed to `Maybe CountOf`. All affected uses of difference of CountOf now handle the Just/Noting cases explicitly or fallback on 'sizeSub' in case that the difference is known to be non-negative from the context. 

Hopefully, my assumption is correct that the difference of two equal `CountOf`s should result in `Just (CountOf 0)`. In fact, the existing implementation of `sizeSub` threw an error in that case, which probably was a bug, wasn't it? This has been corrected as well.

A test for the `isInfixOf` has been added as this function was not covered by tests yet.
Moreover, I think that the function `sub` in Array.Unboxed.Mutable.hs is not covered by test yet, but I was unsure where a test should be added.

Moreover, I wonder if the difference of `Offset`s should also return a `Maybe CountOf` (in Primitive.Types.OffsetSize.hs)? The current implementation ` (Offset a) - (Offset b) = CountOf (a-b)` might actually violate the non-negativity constraint of CountOf. Changing this would likely affect similar amount of places as the changes made so far. What do you think?

Finally, `offsetSub (Offset m) (Offset n) = Offset (m - n)` currently returns `Offset`. Shouldn't that return `CountOf` instead of `Offset`?

### DIFF
--- a/Foundation/Array/Unboxed.hs
+++ b/Foundation/Array/Unboxed.hs
@@ -413,9 +413,8 @@ breakLine :: UArray Word8 -> Either Bool (UArray Word8, UArray Word8)
 breakLine arr@(UArray start len backend)
     | end == start = Left False
     | k2 == end    = Left (k1 /= k2)
-    | k2 == start  = Right (empty, if k2 + 1 == end then empty else unsafeDrop 1 arr)
-    | otherwise    = Right ( unsafeTake (offsetAsSize k1 `sizeSub` offsetAsSize start) arr
-                           , if k2+1 == end then empty else UArray (k2+1) (len `sizeSub` (offsetAsSize (k2+1) `sizeSub` offsetAsSize start)) backend)
+    | otherwise    = let newArray start' len' = if len' == 0 then empty else UArray start' len' backend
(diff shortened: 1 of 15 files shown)


Foundation/Array/Unboxed.hs
```diff
@@ -425,11 +423,11 @@ breakLine arr@(UArray start len backend)
     carriageReturn = 0xd
     goBa ba =
         let k = sysHsMemFindByteBa ba start end lineFeed
-            cr = if k > start then PrimBA.primIndex ba (k `offsetSub` 1) == carriageReturn else False
+            cr = k > start && PrimBA.primIndex ba (k `offsetSub` 1) == carriageReturn
```
vincenthz: :+1: nice !

vincenthz:
Nicely done @ciderale Thanks ! it looks like a nice improvement in places, and in some places the logic is even better than the original.

As a future improvement/research, the only thing that I was thinking was whether moving to a Pattern synonym instead of a boxed Maybe (or optionally a full blown newtype to represent Differential) could be interesting in term of performance/optimisation.

PR #380, opened 2017-07-23
misc improvement to prepare for ghc 8.2 stable

vincenthz:
* Fix warnings with 8.2
* Drop quickcheck/tasty dependency and remove test-foundation stanza
* Port many tests but not all to Check
* drop support for GHC 7.8


vincenthz:
fix #381

PR #386, opened 2017-08-15
Add TypeInType language pragma to Basement.Error

duog:
This was required to compile on ghc head.

Haven't tested it on any other ghcs.

Feel free to close if you don't want this at this time.

### DIFF
--- a/basement/Basement/Error.hs
+++ b/basement/Basement/Error.hs
@@ -3,6 +3,9 @@
 {-# LANGUAGE ImplicitParams #-}
 {-# LANGUAGE ExistentialQuantification #-}
 {-# LANGUAGE CPP #-}
+#if __GLASGOW_HASKELL__ >= 800
+{-# LANGUAGE TypeInType #-}
+#endif
 module Basement.Error
     ( error
     ) where


duog:
Note that __GLASGOW_HASKELL__ has the form Mmm, where M is the major version and mm is the minor version so the test would be > 802. 

This might be a regression, I've opened a ticket here:
https://ghc.haskell.org/trac/ghc/ticket/14121

NicolasDP:
ha my bad: the [CPP doc explains something about that](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/phases.html#standard-cpp-macros).

> For version x.y.z of GHC, the value of __GLASGOW_HASKELL__ is the integer ⟨xyy⟩ (if ⟨y⟩ is a single digit, then a leading zero is added, so for example in version 6.2 of GHC, __GLASGOW_HASKELL__==602). More information in GHC version numbering policy.

Might be worth using the `MIN_VERSION_GLASGOW_HASKELL` as we support [ghc versions >= 7.10](https://github.com/haskell-foundation/foundation/blob/master/foundation.cabal#L33)

duog:
This should not have been accepted without -XTypeInType, as explained in the ticket.

NicolasDP:
Nice, thanks again.

PR #388, opened 2017-08-22
Improve quicksort with alternative partitioning scheme

ciderale:
The current quicksort partition made more swaps and memory access than neccessary.
Especially sorted arrays suffered quite a lot as a benchmark with 10k elements showed.
One improvement is to avoid swaps when the two array positions were actually the same.

The new partitioning uses two cursors to identify elements to be swapped, by scanning from the left and the right, rather than one linear scan from left to right. This reduces the worst-case number of swaps (per partitioning) from n to n/2. Moreover, the new approach avoids multiple reads of the same index position.

The mean execution time from the criterion benchmark are added in the individual commits.
In summary, the runtime reduced from about 18ms to about 9ms on random elements and from over 3 seconds to less than 10milliseconds on pre-sorted data. The latter reduction is certainly also because of using the central element as pivot. However, even without changing the pivot strategy (i.e. using the right most element as pivot) the execution time on sorted arrays reduced to about 39ms.

### DIFF
--- a/basement/Basement/UArray.hs
+++ b/basement/Basement/UArray.hs
@@ -625,9 +625,10 @@ sortBy xford vec
                 p <- partition lo hi
                 qsort lo (pred p)
                 qsort (p+1) hi
-        pivotStrategy lo hi = do
-            pivot <- unsafeRead ma hi
-            pure (hi, pivot)
+        pivotStrategy (Offset low) (Offset high) = do
+            let mid = Offset $ (low + high) `div` 2
+            pivot <- unsafeRead ma mid
+            pure (mid, pivot)
         partition lo hi = do
             (p, pivot) <- pivotStrategy lo hi
             -- RETURN: index of pivot with [<pivot | pivot | >=pivot]
--- a/basement/Basement/UArray.hs
+++ b/basement/Basement/UArray.hs
@@ -625,12 +625,15 @@ sortBy xford vec
                 p <- partition lo hi
                 qsort lo (pred p)
                 qsort (p+1) hi
-        partition lo hi = do
+        pivotStrategy lo hi = do
             pivot <- unsafeRead ma hi
+            pure (hi, pivot)
+        partition lo hi = do
+            (p, pivot) <- pivotStrategy lo hi
             -- RETURN: index of pivot with [<pivot | pivot | >=pivot]
             -- INVARIANT: i & j are valid array indices
-            let go i j = do
-                    -- INVARIANT: k <= pivotpos == hi
+            let go i j pivotpos = do
+                    -- INVARIANT: k <= pivotpos
                     let fw k = do ak <- unsafeRead ma k
                                   if ford ak pivot == LT
                                     then fw (k+1)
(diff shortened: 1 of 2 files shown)


vincenthz:
cool ! I can't process this before the end of the week, but one tip to improve the perf even further is to have the algorithm for the two type of array in `basement/Basement/UArray/BA.hs` and `Addr.hs`.

Using `unsafeRead` and `unsafeWrite` is quite inefficient when having to every time pattern match:

    case arrayType of
             Addr x -> ..
             Native x -> ...


The Addr one should be a copy paste of the BA function provided you use a `Mutable` type just like `Immutable`. Then in UArray you can specialize between the two functions using `onBackendPrim`.

One example would be `filter` which call either UArray/BA.hs filter or UArray/Addr.hs filter.

ciderale:
Thanks for the quick feedback. Having a pattern match on every access is quite expensive, I see.
However, I wonder if it isn't possible to avoid the copy of the algorithm. Isn't it possible to abstract over the two representation?

ciderale:
GHC seems to be quite clever with -O2 optimisation. The call to 'thaw' creates a new array which, if I'm not mistaken, always results in a 'MUArrayMBA' array. With -O2, GHC exploits this and optimises the MUArrayAddr case away. There was no trace of MUArrayAddr left in the core output. However, with -01 the pattern match you mention is there. However, it seems not to be that expensive.

As you suggested, I had a look at avoiding the case expression on the array type in the inner loop.
To avoid duplicating the actual qsort code (when using `onBackendPrim`) I experimented with the following construct.
```
unsafeWithMUArray :: (PrimMonad prim, PrimType ty) 
       => MUArray ty (PrimState prim) 
       -- ^ mutable array to modify 
       -> ((Offset ty -> prim ty) -> (Offset ty -> ty -> prim ()) -> prim a)
       -- ^ create computation based on 'read i' and 'write i value' operations
       -> prim a
unsafeWithMUArray (MUArray start _ (MUArrayMBA (MutableBlock mba))) rwAction 
    = rwAction (\i    -> primMbaRead  mba (start + i)) 
               (\i  v -> primMbaWrite mba (start + i) v)
unsafeWithMUArray (MUArray start _ (MUArrayAddr fptr)) rwAction 
    = withFinalPtr fptr $ \(Ptr addr) -> rwAction (\i   -> primAddrRead  addr (start + i))  
                                                  (\i v -> primAddrWrite addr (start + i) v)
{-# INLINE unsafeWithMUArray #-}
```
It essentially combines unsafeRead/Write and the actual algorithm (e.g. qsort) has to be expressed in terms of `read :: (Offset ty -> prim ty)` and `write :: (Offset ty -> ty -> prim ()`. That is the ` doSort ma = ...` has to be changed as ` doSort unsafeRead unsafeWrite = ...` Hence the distinction between MBA/Addr is done only once and we don't have to manually specialise `doSort` for both cases. What do you think about `unsafeWithMuArray`?

With -O2 optimization both variants show no trace of MUArrayAddr after the optimization. The benchmark execution times slightly differ, but that must be noise, since the core output of the two variants are essentially identical. With -O1, the execution times of the two variants are again quite similar, which is a bit surprising. That suggests that the pattern match is rather inexpensive compared to the actual memory read/write. Or is there something else going on?

The benchmark results for 100k elements:
unsafeRead/Write, -O1
random: mean 468.2 ms (455.9 ms .. 474.9 ms)
sorted: mean 435.1 ms (429.2 ms .. 439.7 ms)

unsafeRead/Write, -O2
random: mean 439.9 ms (436.5 ms .. 442.8 ms)
sorted: mean 423.0 ms (419.7 ms .. 425.4 ms)

unsafeWithMUArray, -01
random: mean 466.9 ms (461.2 ms .. 472.1 ms)
sorted: mean 451.4 ms (446.2 ms .. 454.5 ms)

unsafeWithMUArray, -02
random: mean 432.6 ms (427.3 ms .. 436.9 ms)
sorted: mean 412.9 ms (409.9 ms .. 414.5 ms)

vincenthz:
I've pushed the specialization to both primitives and that seems to make a massive improvement in my benchmark, please have a look to see how it works for you https://github.com/haskell-foundation/foundation/commit/9c70da6c367e550da5c392c5bfa1153681a5b085

vincenthz:
one of the clear downsize of `unsafeWithMUArray` is that the `withFinalPtr` when used on Foreign array would be really costly. Otherwise I'm not sure it's necessary something that is clear enough for GHC to patch all the optimisations strictification&unpacking (it requires looking at core/asm for long).

vincenthz:
I've merged your branch with the specialization I added on top, so that I can move some modules around; we can polish in a further branch/PR the sort performance

ciderale:
Thanks for your additions and merging the PR. 

The `unsafeWithMUArray` was probably a step too far. Yet, I think it is possible to avoid duplicating the actual sort routine and still being performant. Passing `read :: (Offset ty -> prim ty)` and `write :: (Offset ty -> ty -> prim ()` instead of a `MutableByteArray#` or a `Ptr ty` abstracts over the actual representation. This idea generally applies for algorithms that are based on RandomAccess. I will prepare a PR with that idea to discuss the idea on a concrete example.

PR #392, opened 2017-08-27
avoid duplicate inplaceSortBy code

ciderale:
By passing a pair of read & write actions to `inplaceSortBy` the algorithm becomes independent of the underlying representation (Native/Foreign), while still sorting the arrays with specialized code. The benchmark 'types/ByteArray/Sort' showed the same execution time (before and after the change). Hence, we can avoid the duplication of the sort logic. I guess the same idea could be applied to other functions in the `Basement.Alg.*` module. @vincenthz What do you think about that approach?

### DIFF
--- a/basement/Basement/Alg/Foreign/Prim.hs
+++ b/basement/Basement/Alg/Foreign/Prim.hs
@@ -6,7 +6,6 @@ module Basement.Alg.Foreign.Prim
     , primIndex64
     , primRead
     , primWrite
-    , primReadWrite
     ) where
 
 import           GHC.Types
@@ -34,7 +33,3 @@ primRead = primAddrRead
 primWrite :: (PrimMonad prim, PrimType ty) => Mutable (PrimState prim) -> Offset ty -> ty -> prim ()
 primWrite = primAddrWrite
 {-# INLINE primWrite #-}
-
-primReadWrite :: (PrimMonad prim, PrimType ty) => Mutable (PrimState prim) -> ((Offset ty -> prim ty), (Offset ty -> ty -> prim ()))
-primReadWrite addr = (primAddrRead addr, primAddrWrite addr)
-{-# INLINE primReadWrite #-}
--- a/basement/Basement/Alg/Foreign/Prim.hs
+++ b/basement/Basement/Alg/Foreign/Prim.hs
@@ -6,6 +6,7 @@ module Basement.Alg.Foreign.Prim
     , primIndex64
     , primRead
     , primWrite
+    , primReadWrite
     ) where
 
 import           GHC.Types
@@ -33,3 +34,7 @@ primRead = primAddrRead
 primWrite :: (PrimMonad prim, PrimType ty) => Mutable (PrimState prim) -> Offset ty -> ty -> prim ()
 primWrite = primAddrWrite
 {-# INLINE primWrite #-}
+
+primReadWrite :: (PrimMonad prim, PrimType ty) => Mutable (primState prim) -> ((Offset ty -> prim ty), (Offset ty -> ty -> prim ()))
+primReadWrite addr = (primAddrRead addr, primAddrWrite addr)
+{-# INLINE primReadWrite #-}
--- a/basement/Basement/Alg/Foreign/Prim.hs
+++ b/basement/Basement/Alg/Foreign/Prim.hs
@@ -35,6 +35,6 @@ primWrite :: (PrimMonad prim, PrimType ty) => Mutable (PrimState prim) -> Offset
 primWrite = primAddrWrite
 {-# INLINE primWrite #-}
 
-primReadWrite :: (PrimMonad prim, PrimType ty) => Mutable (primState prim) -> ((Offset ty -> prim ty), (Offset ty -> ty -> prim ()))
+primReadWrite :: (PrimMonad prim, PrimType ty) => Mutable (PrimState prim) -> ((Offset ty -> prim ty), (Offset ty -> ty -> prim ()))
 primReadWrite addr = (primAddrRead addr, primAddrWrite addr)
 {-# INLINE primReadWrite #-}
--- a/basement/Basement/Alg/Foreign/PrimArray.hs
+++ b/basement/Basement/Alg/Foreign/PrimArray.hs
@@ -15,7 +15,6 @@ module Basement.Alg.Foreign.PrimArray
     , any
     , filter
     , primIndex
-    , inplaceSortBy
     ) where
 
 import           GHC.Types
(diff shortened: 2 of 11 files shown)


vincenthz:
I think this is only workable when the type is specialized, otherwise a quick benchmark tell me that it this approach goes slower when ghc hasn't specialized the prim function to a specific type (here with array of Word32), which are very likely the result of further boxes (tuple, function not resolved at compilation,etc):

before:

```
benchmarking types/ByteArray/Sort32/bs2000/UArray_W32
time                 984.3 μs   (980.6 μs .. 987.7 μs)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 983.3 μs   (979.4 μs .. 996.6 μs)
std dev              21.76 μs   (5.438 μs .. 54.15 μs)
variance introduced by outliers: 12% (moderately inflated)
```

after:

```
benchmarking types/ByteArray/Sort32/bs2000/UArray_W32
time                 1.105 ms   (1.100 ms .. 1.110 ms)
                     1.000 R²   (1.000 R² .. 1.000 R²)
mean                 1.103 ms   (1.101 ms .. 1.107 ms)
std dev              8.788 μs   (5.793 μs .. 13.49 μs)
```

ciderale:
Okay, I see. You are right, there is some overhead involved. I will experiment with the  'types/ByteArray/Sort32/bs2000/UArray_W32' benchmark to see if it is possible to overcome the problem. I still hope that the code duplication can be avoided.

ndmitchell:
One trick to consider is using the static argument transformation on the two input arguments, as an example:

```haskell
{-# INLINE sortBy #-}
inplaceSortBy unsafeRead unsafeWrite = myInplaceSortBy
    where
        myInplaceSortBy ford start len = ... using unsafeRead ...
```

I suggest putting the two unsafeRead/write as the first two arguments, and not having the tuple, since that will definitely improve GHC's chances of matching without duplication (I think the duplication is really yuk...)

ciderale:
Thanks for the input. I was currently experimenting with a typeclass:
```
class RandomAccess a ty where
    read  :: PrimMonad prim => a ty (PrimState prim) -> (Offset ty)       -> prim ty
    write :: PrimMonad prim => a ty (PrimState prim) -> (Offset ty) -> ty -> prim ()

inplaceSortBy :: (PrimMonad prim, RandomAccess a ty) 
              => (ty -> ty -> Ordering) -> (Offset ty) -> (CountOf ty)
              -> a ty (PrimState prim) -- ^ Data to be sorted
              -> prim ()

instance PrimType ty => Algorithm.RandomAccess MutableBlock ty where
    read (MutableBlock mba) = primMbaRead mba
    write (MutableBlock mba) = primMbaWrite mba

data PtrSt ty st = PtrSt Addr#
instance PrimType ty => Algorithm.RandomAccess PtrSt ty where
    read (PtrSt addr) = PrimAddr.primRead addr
    write (PtrSt addr) = PrimAddr.primWrite addr
```
For UArray the benchmark looks good and the core output (before/after) are identical (modulo variable renaming). 

I'm a bit unsure if this is too heavy machinery, since I had to enable MultiParamTypeClasses and FlexibleContexts. Especially FlexibleContexts is something that I'm not really familiar with. So any feedback on the above (including naming) are very appreciated.

Excluding the `PrimType ty` constraint from the class definition was a deliberate choice. This allows to reuse `inplaceSortBy` for `BoxedArray` types, thus reducing one more duplicate of the sorting routine.

ndmitchell:
Your typeclass extensions are positively pedestrian - on some projects I have them on by default. For reference, I would have gone with:

```
class RandomAccess container prim ty where
    read  :: container -> Offset ty   -> prim ty
    write :: container -> Offset ty -> ty -> prim ()
```

The requirement that `a` have the shape it does seems very artificial - you couldn't use it for `MutableBoolVector` on top of `Integer`, for example. You also hoist the `PrimMonad` class one level up, which seems cleaner.

ciderale:
Thanks for the clarification and the cleaned-up class. I will make the necessary changes to integrate this.

Regarding your last comment about the static argument transformation. That was already done
```
inplaceSortBy ford start len (unsafeRead, unsafeWrite) 
    = qsort start (start `offsetPlusE` len `offsetSub` 1) where qsort ... use unsafeRead ...
```
so I could only try avoiding the tuple.

ndmitchell:
Avoiding the tuple is a good idea. But so is moving all arguments that don't change "that much" further up in the list - e.g. `unsafeRead` never changes but `len` often does. You might also get benefits from having the two lambdas - so the first only takes `unsafeRead`/`write`, and a subsequent lambda gets the changing arguments.

ciderale:
The typeclass based approach does, as far as I can tell, not incur the overhead of the previous approach based on tuples. Simply changing tuples to individual arguments didn't show any improvements. Based on the typeclass abstraction, it was also possible to change BoxedArray.sortBy to use Algorithm.inplaceSortBy. 

The typeclass definition `class RandomAccess container prim ty` (rather the one in the my earlier comment) was already helpful as `Ptr ty` (for the Foreign backend) is not parametrized by prim. 

Below are the benchmark results on Word32 data. The result for UArray and Block are comparable.
The significant gain on Array_W32 is due to the fact of using the new partitioning scheme for quicksort that was merged in PR #390.

I'm curious what you think about the proposed changes.

Benchmarks original:
```
benchmarking types/ByteArray/Sort32/2000/Array_W32
time                 1.681 ms   (1.651 ms .. 1.714 ms)
                     0.995 R²   (0.990 R² .. 0.998 R²)
mean                 1.721 ms   (1.693 ms .. 1.759 ms)
std dev              112.5 μs   (92.46 μs .. 137.1 μs)
variance introduced by outliers: 48% (moderately inflated)

benchmarking types/ByteArray/Sort32/2000/UArray_W32
time                 1.007 ms   (987.3 μs .. 1.023 ms)
                     0.997 R²   (0.995 R² .. 0.998 R²)
mean                 1.004 ms   (991.0 μs .. 1.021 ms)
std dev              54.79 μs   (43.52 μs .. 68.11 μs)
variance introduced by outliers: 44% (moderately inflated)

benchmarking types/ByteArray/Sort32/2000/Block_W32
time                 1.018 ms   (989.8 μs .. 1.051 ms)
                     0.994 R²   (0.990 R² .. 0.998 R²)
mean                 1.009 ms   (994.1 μs .. 1.028 ms)
std dev              53.42 μs   (43.56 μs .. 68.92 μs)
variance introduced by outliers: 43% (moderately inflated)
```

Benchmark after:
```
benchmarking types/ByteArray/Sort32/2000/Array_W32
time                 244.1 μs   (239.3 μs .. 250.6 μs)
                     0.992 R²   (0.983 R² .. 0.997 R²)
mean                 258.3 μs   (252.5 μs .. 270.9 μs)
std dev              29.10 μs   (14.78 μs .. 53.22 μs)
variance introduced by outliers: 83% (severely inflated)

benchmarking types/ByteArray/Sort32/2000/UArray_W32
time                 1.005 ms   (993.6 μs .. 1.017 ms)
                     0.998 R²   (0.998 R² .. 0.999 R²)
mean                 1.003 ms   (992.2 μs .. 1.018 ms)
std dev              46.29 μs   (34.90 μs .. 62.15 μs)
variance introduced by outliers: 37% (moderately inflated)

benchmarking types/ByteArray/Sort32/2000/Block_W32
time                 1.007 ms   (989.5 μs .. 1.031 ms)
                     0.997 R²   (0.994 R² .. 0.999 R²)
mean                 995.4 μs   (985.1 μs .. 1.013 ms)
std dev              45.53 μs   (33.33 μs .. 64.35 μs)
variance introduced by outliers: 35% (moderately inflated)
```

ciderale:
I'm not sure what happened with the failing 'GHC-8/OSX' build. The output does not tell much. Is that a problem of my commits or is it a general build problem?

vincenthz:
@ciderale occasionally Travis build are not reliable. I've kicked it back, and hopefully will go through now

vincenthz:
Ok I think that make sense, and we should go in this direction; although before proceeding I'll check further benchmarks / core dump to make sure there's no surprise

ciderale:
Cool! Let's hope there is no suprise there. Let me know what your conclusions are. We then should also introduce a similar typeclass for immutable data, which covers even more of the algorithms.

PR #395, opened 2017-08-31
Adjust includes in foundation_random.c to work with both libc and musl

### LINKED ISSUE #393: Doesn't currently build against musl
Musl - a popular safer `libc` variant used by Alpine Linux and others - has `types.h`, but not `linux/types.h`, so builds of Foundation fail on musl-based systems like Alpine. Would be great to be able to build Foundation there too, especially since Alpine has GHC 8 now, and Stack compiles there happily. Unsure whether the same applies to `linux/random.h`.

purcell:
This fixes #393

Test process was to set Stack resolver to lts-9.2, then build with "--docker", thereby using the default (libc) fpco linux image. Then the build was repeated using a musl-based docker image produced with the following Dockerfile:

FROM alpine:latest
RUN apk add ghc musl-dev shadow

Both builds succeeded.

### DIFF
--- a/cbits/foundation_random.c
+++ b/cbits/foundation_random.c
@@ -10,8 +10,7 @@
 
 #if defined(FOUNDATION_SYSTEM_LINUX)
 #include <sys/syscall.h>
-#include <linux/types.h>
-#include <linux/random.h>
+#include <sys/types.h>
 #include <unistd.h>
 #ifndef _GNU_SOURCE
 #define _GNU_SOURCE


purcell:
Does this look reasonable? I'd hope it's fairly uncontroversial. :-)

vincenthz:
sorry about that, I didn't have time to test that everything is still as it should be yet; if you don't have the right includes on linux, then the random syscall is not available, but it gracefully fallback to the previous behavior using /dev/urandom. so while the build is a success (which is great), it doesn't necessarily mean that nothing is broken

purcell:
Aha, good to know. I'm not familiar with what would constitute an adequate `random` by Foundation's standards.

Are there any Linux `libc` implementations other than `glibc` that are supported? If not, I can modify this PR to use the original includes if `__GLIBC__` is defined.

vincenthz:
I've checked and this work as advertised ! thanks for hanging on !

purcell:
Awesome - thanks for your guidance.

PR #398, opened 2017-09-05
Touch the bytearray after the action in uarray's withPtr when pinned

### LINKED ISSUE #397: withPtr is unsafe (maybe?)
Looking at

https://github.com/haskell-foundation/foundation/blob/08dd88a98ed1eb793e01423aee6f898da9defe3c/basement/Basement/UArray.hs#L295:

This takes an array addressed by a pinned bytearray, and converts it to a pointer. But what keeps the pinned bytearray alive afterwards? I can't see anything, and I have a bug in an application of mine that looks suspiciously like random memory corruption from exactly this problem.

Furthermore, I note that a few lines down you copy an unpinned array to a pinned array and pass it off to withPtr. After that you touch the pinned array - I think that would be unnecessary if withPtr did it's job properly.

https://github.com/haskell-foundation/foundation/blob/08dd88a98ed1eb793e01423aee6f898da9defe3c/basement/Basement/UArray.hs#L303-L304

vincenthz:
After the bytearrayContent#, there's potentially nothing holding
the bytearray, so the GC could in rare race decide to get rid of it
before its address is used by f.

fix #397

### DIFF
--- a/Foundation/Time/StopWatch.hs
+++ b/Foundation/Time/StopWatch.hs
@@ -58,7 +58,7 @@ initPrecise = unsafePerformIO $ do
     let p32 = castPtr p :: Ptr Word32
     !n <- peek (p32 `ptrPlus` ofs_MachTimebaseInfo_numer)
     !d <- peek (p32 `ptrPlus` ofs_MachTimebaseInfo_denom)
-    -- touch mti ..
+    mutableTouch mti
     pure (integralUpsize n, integralUpsize d)
 {-# NOINLINE initPrecise #-}
 #endif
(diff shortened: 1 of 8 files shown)


ndmitchell:
Note the race isn't super rare. If you do something "meaty" inside the `withPtr` you can lose the race all the time. Even something like an `hPut` will lose the race 0.1% of times.

ndmitchell:
Why is StopWatch not using the with variant? It seems ideally suited to it.

vincenthz:
it's only using a pointer pervasively because the access primitive for the bytearray are missing quite a few cases. the peek is only "temporary" until I have everything I need to access through the bytearray# instead of the Addr.

PR #402, opened 2017-09-09
Extended Words

vincenthz:


NicolasDP:
would be nice to have the storable instances too

PR #409, opened 2017-09-16
Compatibility with the Semigroup/Monoid proposal

bgamari:


vincenthz:
Thanks for doing this; To make it less painful, using the Basement.Compat.<Class> approach for Semigroup just like other classes would remove most ifder'y from the high level modules

bgamari:
Vincent Hanquez <notifications@github.com> writes:

> Thanks for doing this; To make it less painful, using the
> Basement.Compat.<Class> approach for Semigroup just like other classes
> would remove most ifder'y from the high level modules
>
Sure, I'll rework this accordingly.

PR #411, opened 2017-09-22
Exceptions

vincenthz:
adding some MonadBracket and exceptions stuff, and add the ExceptT transformers

(first time using MonadBracket abstractions,  review feedback from a senior exceptions tamer @snoyberg would be useful :) )


vincenthz:
Thanks for the review !

PR #412, opened 2017-09-26
test-suite: depend on template-haskell

domenkozar:
Tests seems to use TH:

```
$ git grep TemplateHaskell
tests/Test/Foundation/Misc.hs:{-# LANGUAGE TemplateHaskell #-}
```

Otherwise darwin compilation fails with:

```
Linking dist/build/check-foundation/check-foundation ...
running tests
Running 1 test suites...
Test suite check-foundation: RUNNING...
dyld: Library not loaded: @rpath/libHStemplate-haskell-2.11.1.0-ghc8.0.2.dylib
  Referenced from: /private/tmp/nix-build-foundation-0.0.15.drv-0/foundation-0.0.15/dist/build/check-foundation/check-foundation
  Reason: image not found
Test suite check-foundation: FAIL
Test suite logged to: dist/test/foundation-0.0.15-check-foundation.log
0 of 1 test suites (0 of 1 test cases) passed.
builder for ‘/nix/store/hm5hdkapi57ri0bv3vrr5xmq4mll2gns-foundation-0.0.15.drv’ failed with exit code 1
```

### DIFF
--- a/foundation.cabal
+++ b/foundation.cabal
@@ -235,6 +235,7 @@ test-suite check-foundation
   build-depends:     base >= 3 && < 5
                    , basement
                    , foundation
+                   , template-haskell
   ghc-options:       -Wall -fno-warn-orphans -fno-warn-missing-signatures
   default-language:  Haskell2010
   if impl(ghc >= 8.0)


vincenthz:
I'ld rather see a conditional than introduce a hard dependency. we strictly use this *only* for appveyor making sure things works properly on windows.

domenkozar:
Forgot to mention, this only happens on macOS, but not on Linux. I'd assume you should be able to reproduce there even with Stack.

domenkozar:
`stack test` passes on macOS otherwise, it seems that it only fails with Nix+macOS.

PR #417, opened 2017-10-11
Avoid touch, use with, fixes #399

### LINKED ISSUE #399: Stop using byteArrayContents# directly
Following #397, I read through the code base looking for `byteArrayContents#`. It's used incorrectly every time. I recommend defining `withByteArrayPtr :: ByteArray# -> (Ptr a -> IO b) -> IO b` or similar to capture the correct pattern with the required `touch`. In particular:

https://github.com/haskell-foundation/foundation/blob/dc8d4f7c1ffb34878781c845387835ca7edb4443/basement/Basement/UArray/Mutable.hs#L127

This looks like another instance of #397, almost verbatim.

https://github.com/haskell-foundation/foundation/blob/dc8d4f7c1ffb34878781c845387835ca7edb4443/basement/Basement/Block/Mutable.hs#L87-L88

This defines a function `mutableGetAddr` which is just as unsafe as `byteArrayContents#` which is then used unsafely in StopWatch. In particular I loved:

https://github.com/haskell-foundation/foundation/blob/dc8d4f7c1ffb34878781c845387835ca7edb4443/Foundation/Time/StopWatch.hs#L55-L61

The touch is there, just as a comment...

ndmitchell:
The main purpose of this change is to avoid the do something unsafe and touch pattern, and instead move to using `withFoo`, even for the low-level pieces. In the process I think it fixes bugs in Stopwatch, but mostly it's about static guarantees in more places.

At the same time I changed onBacked to use `Block ty` rathern than `ByteArrayContents#` as this lets you use the `with` functions on safer types, and also preserves type information for longer. It also provides a nice symmetry with the `Ptr` - a newtype phantom type over `Addr#`.

### DIFF
--- a/Foundation/Foreign/MemoryMap/Windows.hs
+++ b/Foundation/Foreign/MemoryMap/Windows.hs
@@ -9,7 +9,6 @@ import Control.Exception hiding (handle)
 
 import Basement.Compat.Base
 import Basement.Types.OffsetSize
-import Basement.FinalPtr
 import Foundation.VFS
 import Foundation.Foreign.MemoryMap.Types
 
--- a/Foundation/Time/StopWatch.hs
+++ b/Foundation/Time/StopWatch.hs
@@ -53,7 +53,7 @@ initPrecise = unsafePerformIO $ integralDownsize <$> queryPerformanceFrequency
 initPrecise :: (Word64, Word64)
 initPrecise = unsafePerformIO $ do
     mti <- newPinned (sizeOfCSize size_MachTimebaseInfo)
-    mutableWithAddr mti $ \p -> do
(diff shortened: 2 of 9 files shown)


vincenthz:
Looking good :+1: . the refactor of backend function to get a MutableBlock/Block instead of a primitive was eventually planned, so nice to see it there (albeit making review a bit more involved).

The only minor thing, is that I would have kept the touch / mutableTouch function as internal helpers instead of inlining them

ndmitchell:
Thanks for the review - I didn't originally intend to do the onBackend stuff but it turned out to be required halfway through. I followed up with the requested changes in #419.

PR #420, opened 2017-10-19
Use CApiFFI for variadic fcntl 

angerman:
This came up when using the alternative `-llvmng` backend. Which started to complain about two functions `fcntl` that had different signatures. While I assume full responsibility for all llvmng bugs, it was suggested by @geekosaur, to rather use the CApiFFI extension to import variadic `fcntl` as multiple functions with different numbers of arguments.

### DIFF
--- a/Foundation/System/Bindings/Posix.hsc
+++ b/Foundation/System/Bindings/Posix.hsc
@@ -17,6 +17,7 @@ module Foundation.System.Bindings.Posix
 
 import Basement.Compat.Base
 import Foreign.C.Types
+import Foreign.Ptr (nullPtr)
 import Data.Bits
 import Foundation.System.Bindings.PosixDef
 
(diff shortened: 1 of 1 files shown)


vincenthz:
I'm pretty sure that POSIX doesn't say that a variadic function is equivalent to a function of N parameters with the defaulted value to be 0.

The only truly valid function is to consider it a truly variadic function (through CAPI, which AFAIK is the only ghc calling convention capable of handling variadic), or not to care that much, since it's not really used variadically, i.e. no one is using va_args on the other side (on most system foundation care about I think, althought haven't checked windows compat layer), the second argument determine by itself the number of actual useful argument, the register will always have a value, but I see the init to zero as the wrong solution to this problem.

There's really a need to have multiple functions I think, defaulting to 0 is not a valid solution (i.e. you push the callee the idea that you have 3 arguments when you expect 2, if you'ld end using the va_arg stuff). Also just for the sake of avoiding to convert an int into a pointer in library code, it's useful to have different bindings for the `int parameter` and the `ptr parameter`

angerman:
@vincenthz thank you for taking a look at it.

As this PR now essentially replaces the ccall FFI
```haskell
foreign import ccall unsafe "fcntl"
```
with the capi FFI call
```haskell
foreign import capi "fcntl.h fcntl"
```
does it have any chance of being merged?

I'll adjust the PR subject and summary.

vincenthz:
yes, I want to read about capi in more detail first though.

angerman:
Did you have any chance to look at the capi yet?

vincenthz:
so I'm not convinced that I know what I wanted to know about it , but it's seems quite harmless considering it's use all over the place in base

vincenthz:
Thanks !

PR #434, opened 2017-11-01
Foreign/Native-PrimArray Algorithms Code-Deduplication

ciderale:
This PR combines the Foreign/Native implementation for the Algorithms in PrimArray.hs in a similar way as done in #414 . I also added an inline pragma for `filter`: it was the only function in PrimArray and I suspect the pragma was missing (or was that on purpose?).

The benchmark look similar and the core-output for UArray.hs/Block.hs is almost identical. I found `vimdiff` quite effective to visually highlight the difference between two versions of core -- after sed'ing a bindings "random" suffix away). The differences (to master) where: 
- the inlining of filter (of course)
- a flip of operations on calls of withUnsafeFinalPtr (see below)
- let expressions (related to length or Eq dictionary) are floated towards the top-level

Example-Fragment Before:
```code. 
withUnsafeFinalPtr  $fPrimMonadST fptr
   ((\ (x :: Ptr ty) (s1 :: State# Any) ->                 
           (# s1,  case x of { Ptr addr ->                          
               let {  $dEq :: Eq ty                                  
                        $dEq = $p1PrimType w } in   joinrec {   ...
```

Example-Fragment After:
```code. After
let {  $dEq :: Eq ty             -- moved upwards                                        
         $dEq = $p1PrimType w } in   
...
withUnsafeFinalPtr $fPrimMonadST fptr
           ((\ (ptr :: Ptr ty) (eta :: State# Any) ->              
                  case ptr of { Ptr ds3 ->            -- case & tuple interchanged       
                        (# eta,      joinrec {  ...  
```code


vincenthz:
This is looking great (sorry about the slow review). The only things that I'm not quite sure if the `Maybe`s; I avoided them on purpose to prevent the creation of a box of `a` when we are explicitely working with indexes and we have a known value as a sentinel (`end`). I imagine that having them completely unboxed would fuse better (go faster) to any composite operation, at the cost of losing the safety of having to check for algorithm's end condition.

vincenthz:
I merged up to 252f8b562a2bc9cbad73e59aa89b564c51404a5e (nice work again !) and we can discuss the last 2 commits

ciderale:
Thanks for merging the first part. no worries for the time delay I was pretty busy these days too. 

Concerning the maybes. With the currently merged version the ‘end’ is computed twice. Once by the onBackendPure’ and once at the call site. Using maybe avoids this double computation, at the potential cost of an allocation. I don’t know what is worse. 

Moreover, If I remember correctly, the maybe is optimised away in the native case (when an operation follows the find index). I have to double check that tough. The maybe was only allocated in the foreignptr case. Probably because there are too many layers of ‘withPtr’ separating the construction and deconstruction of the maybe, so ghc is not optimising it away.  

I was thinking about adding a continuation function to onBackendPure’. This should bring the construction and deconstruction of the maybe closer together and ghc should be able to optimise it again. What do you think about this idea?

Ps: I was not sure, if any of the benchmarks are actually using foreignptr. 
I have the impression that they all use native memory. But I didn’t see how I could create a uarray backed by foreignptr. You certainly can help me on that point. 

Von meinem iPhone gesendet

ciderale:
I think we could also use '-1' as sentinel (instead of the `end` index). That way we could also avoid computing the 'end' twice. Or is there a case where '-1' would be a valid `Offset ty`?

Using '-1' as sentinel would be a bit more explicit than `end` I think, but still less explicit than using `Maybe`s. For this low-level tasks, I agree that performance is key. But, it would still be nice to use `Maybe` since that is what `Maybe`s are for. I could check if a continuation passing style helps GHC to eliminate the Maybe. @vincenthz What do you think about it?

Issue #433, opened 2017-11-01
Add terminal size (row+column) capability in Basement.Terminal


vincenthz:
termios.h `TIOCGWINSZ` ioctl on unix:

```
    struct winsize max;
    ioctl(0, TIOCGWINSZ , &max);
    printf ("lines %d\n", max.ws_row);
    printf ("columns %d\n", max.ws_col);
```

On Windows:

```    CONSOLE_SCREEN_BUFFER_INFO csbi;
    int columns, rows;

    GetConsoleScreenBufferInfo(GetStdHandle(STD_OUTPUT_HANDLE), &csbi);
    columns = csbi.srWindow.Right - csbi.srWindow.Left + 1;
    rows = csbi.srWindow.Bottom - csbi.srWindow.Top + 1;
```

A python example on how to achieve that:
https://gist.github.com/jtriley/1108174

andymac-2:
I was hoping to try this one this weekend, can somebody let me know the CPP macros that will be defined when the target OS is windows, linux or other?

PR #438, opened 2017-11-04
Add terminal size (row+column) capability in Basement.Terminal

andymac-2:
reference #433 
Tested on Windows, OSX.
Let me know if this is the sort of thing you're looking for.


vincenthz:
@andymac-2 thanks for working on this ! I was more thinking of a haskell methods using the bindings to C function (ioctl) and the System.Win32 modules, than some C functions

andymac-2:
No worries. It seemed a bit weird to just use those calls once. What I could do is some boilerplate to expose some parts of libc and POSIX C first. However my expertise is not in windows. I feel as that may be useful for other applications. Any objections to me writing a `Basement.C.Libc` or `Basement.C.POSIX` or similar?

Foundation/System/Bindings/POSIX/Stropts.hsc
```diff
@@ -0,0 +1,235 @@
...
+foreign import ccall "ioctl_winsize" c_ioctl_winsize :: CInt -> Ptr Winsize -> IO CInt
+#elif defined FOUNDATION_SYSTEM_WINDOWS
+foreign import ccall "GetStdHandle" c_get_std_handle :: CULong -> IO Handle
+foreign import ccall "GetConsoleScreenBufferInfo" c_get_console_screen_buffer_info :: Handle -> Ptr ConsoleScreenBufferInfo -> IO CInt
+#endif
+
+#ifdef FOUNDATION_SYSTEM_UNIX
+ioctlWinsize :: CInt -> IO (Maybe (Int, Int))
+ioctlWinsize fd = do
+    winsizePtr <- malloc :: IO (Ptr Winsize)
```
vincenthz: here you can use `alloca` or `allocaBytes` instead of `malloc` + `free`.

Foundation/System/Bindings/POSIX/Stropts.hsc
```diff
@@ -0,0 +1,235 @@
...
+-- defined FOUNDATION_SYSTEM_WINDOWS
+
+
+foreign import ccall "popen" c_popen :: CString -> CString -> IO (Ptr CFile)
+foreign import ccall "pclose" c_pclose :: Ptr CFile -> IO CInt
+foreign import ccall "fscanf_int" c_fscanf_int :: Ptr CFile -> Ptr CInt -> IO CInt
+#ifdef FOUNDATION_SYSTEM_UNIX
+foreign import ccall "ioctl_winsize" c_ioctl_winsize :: CInt -> Ptr Winsize -> IO CInt
+#elif defined FOUNDATION_SYSTEM_WINDOWS
+foreign import ccall "GetStdHandle" c_get_std_handle :: CULong -> IO Handle
```
vincenthz: I think this should be stdcall for windows 32 bits here.
also I'm surprise this is not available in `win32` already, I think GetStdHandle is available through `Graphics.Win32.Misc`, but I think it would probably to contribute GetConsoleScreenBufferInfo in `win32` too !

Foundation/System/Bindings/POSIX/Stropts.hsc
```diff
@@ -0,0 +1,235 @@
...
+                    return (width height)
+#endif
+-- defined FOUNDATION_SYSTEM_WINDOWS
+            
+if' :: Bool -> a -> a -> a
+if' True t _ = t
+if' False _ f = f
+
+popen :: String -> String -> IO (Ptr CFile)
+popen cmd_str mode_str = do
```
vincenthz: I don't think we need any popen thing; Personally I'm happy if we just return a constant (80x25) when the first method of gathering information fails. `popen` doesn't really play nice with the runtime and it's doing a lots of work (fork, exec, setting up pipe, read/write) which are blocking and assume lots of things

Foundation/System/Bindings/POSIX/Stropts.hsc
```diff
@@ -0,0 +1,235 @@
...
+#elif defined FOUNDATION_SYSTEM_UNIX
+    maybeWinsize <- ioctlWinsize 0
+#endif
+    case maybeWinsize of
+         Just winsize1 -> return winsize1
+         Nothing -> do
+             maybeWinsize2 <- tputWinsize
+             case maybeWinsize2 of
+                  Just winsize2 -> return winsize2
+                  Nothing -> return (80, 25)
```
vincenthz: :+1:

foundation.cabal
```diff
@@ -111,6 +111,7 @@ library
                      Foundation.UUID
                      Foundation.System.Entropy
                      Foundation.System.Bindings
+                     Foundation.System.Bindings.POSIX.Stropts
```
vincenthz: I would rather see this still in basement. I appreciate there's a bits of a complication with the dependencies, but I'ld the basic terminal stuff as low as possible to get away with building a good abstraction directly in basement (which need to do some terminal stuff, e.g. `error`)

vincenthz:
awesome tedious work ! Try to make sure you don't commit random by-product in your PR, otherwise I'll have to clean it up on my side. I was really under the impression we could leverage win32 a bit more in this case, but apparently that's not the case

vincenthz:
also forgot to say, but there should be a way not to duplicate `foundation_system.h`. maybe fudging with the path on the C-includes

andymac-2:
Yeah, things are all over the place right now while I try to debug and test. I haven't used git in a long time, so I had no idea my commits would end up here. I'll make a branch and clean it up. The copy of system foundation was just a quick fudge while I debug, but I can include the regular path. I was trying not to use dependencies which aren't part of base, otherwise there are several improvements I would have made using `win32` `System.Command` and some monad libraries.

PR #439, opened 2017-11-06
fixed examples haddocks; added second example

GinBaby:
Hello again, I think I messed up the haddocks for the Examples headers, so I fixed that and added an example for `second.` 

Also fixed my text editor to eliminate trailing white spaces. Thank you for your patience. :)

### DIFF
--- a/basement/Basement/Compat/Bifunctor.hs
+++ b/basement/Basement/Compat/Bifunctor.hs
@@ -5,19 +5,19 @@
 -- Stability   : experimental
 -- Portability : portable
 --
--- A bifunctor is a type constructor that takes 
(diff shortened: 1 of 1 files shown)


vincenthz:
Just noticed, but I'm not sure that make sense to me that `Bifunctor` is a `type constructor`. Maybe a `type construction` would be better ? Otherwise it's looking good !

vincenthz:
merging since it fixes a parsing issue in the comment anyhow

GinBaby:
Hi, sorry, I'm right in the middle of moving so I put off replying to you until I got my things moved into the new place. I'll think about the wording and open a new PR with some revisions soon. Cheers :)

Issue #440, opened 2017-11-06
basement doesn't compile with GHC-7.8.4


phadej:
But its bounds advertise they would. Please either test with GHC-7.8.4 or explicitly drop the support, i.e. adjust lower-bound.

phadej:
`foundation-0.17` doesn't compile with GHC-7.8 either. It seems HVR made a revision already. 

http://hackage.haskell.org/package/foundation-0.0.17/revisions/

PR #443, opened 2017-11-11
Add ghc-7.8 travis job

phadej:


phadej:
```
/home/travis/build/haskell-foundation/foundation/Foundation/Exception.hs:12:31:
        Could not deduce (Applicative m) arising from a use of ‘pure’
        from the context (MonadBracket m)
          bound by the type signature for
```

is legit bug, which I have no time to fix. Please add this job to the travis, so you can fix #440 .

gwils:
I have a fix for that bug in #460

vincenthz:
merging along #460, Thanks

Issue #429, opened 2017-10-30
Class instances for types from Foreign.C.Types


leohaskell:
Why does only the [CSize](https://hackage.haskell.org/package/base-4.10.0.0/docs/Foreign-C-Types.html#t:CSize) from  [Foreign.C.Types](https://hackage.haskell.org/package/base-4.10.0.0/docs/Foreign-C-Types.html) has instances for *Additive*, etc.?
If there is not a reason for that, I'm willing to implement those instances for all types from [Foreign.C.Types](https://hackage.haskell.org/package/base-4.10.0.0/docs/Foreign-C-Types.html).
If that's alright, perhaps these instances should be in modules **Basement.FFI.Additive**, etc. and **Foundation.FFI** (or **C.Types** instead of **FFI**)?

Note: I think that at some stage, it would be nice to implement the things from [Foreign.C.String](https://hackage.haskell.org/package/base-4.10.0.0/docs/Foreign-C-String.html) using the *Foundation*'s *String*.

vincenthz:
I can't think of a reason why it's not implemented yet.  PR welcomed.

I don't think you'll be able to have those instance in anywhere else than Basement.Numerical.* since otherwise you'll hit the OrphanInstance problem.

Also for Foreign.C.String, It sounds like a good idea to me !

PR #444, opened 2017-11-14
Implement Numerical class instances for the foreign c types

leohaskell:
See [this issue](https://github.com/haskell-foundation/foundation/issues/429).


leohaskell:
I have to say, I'm not really happy with the module structure, but I didn't know where to put the C types (now they're in `Basement.Compat.C.Types`).

vincenthz:
related to CBool, to be in line with what basement is doing, the right thing to do with to provide CBool ourself (see Basement.Compat.Natural for example) if it doesn't exist. this prevent a bunch of ifdef in the instances and provide a better module for users.

Otherwise, thanks for adding the C instances

PR #451, opened 2017-12-21
Use isTrue# directly

ndmitchell:
Mostly fixing this because I went looking at foundation for code to make this change, ripped it out of foundation, and then later found there is a better way to do it.


vincenthz:
Thanks !

PR #455, opened 2018-01-07
Add better implementation of Builder

NicolasDP:
* add Builder for `Block Word8`
* add Builder for `String`

also introduce `Cast` type family in basement instead of `From` for certain types.

one of the issue with `From` is that it kinda implies _value conversion_ while actually we were litteraly casting the value into the new type:

example:

```haskell
cast (-10 :: Int) :: Word == 18446744073709551606
```

### DIFF
--- a/basement/Basement/Alg/Native/UTF8.hs
+++ b/basement/Basement/Alg/Native/UTF8.hs
@@ -25,7 +25,6 @@ module Basement.Alg.Native.UTF8
     , primIndex64
     , primRead8
     , primWrite8
-    , sizeChar
     ) where
 
 import           GHC.Int
@@ -125,16 +124,6 @@ prevSkip ba offset = loop (offset `offsetMinusE` sz1)
     loop o
         | isContinuation (primIndex8 ba o) = loop (o `offsetMinusE` sz1)
         | otherwise                       = o
-
-sizeChar :: Char -> CountOf Word8
-sizeChar !c
-    | bool# (ltWord# x 0x80##   ) = CountOf 1
-    | bool# (ltWord# x 0x800##  ) = CountOf 2
-    | bool# (ltWord# x 0x10000##) = CountOf 3
-    | otherwise                   = CountOf 4
-  where
-    !(I# xi) = fromEnum c
-    !x       = int2Word# xi
 
 write :: PrimMonad prim => Mutable (PrimState prim) -> Offset8 -> Char -> prim Offset8
 write mba !i !c
--- a/basement/Basement/Alg/Native/UTF8.hs
+++ b/basement/Basement/Alg/Native/UTF8.hs
@@ -25,6 +25,7 @@ module Basement.Alg.Native.UTF8
     , primIndex64
     , primRead8
     , primWrite8
+    , sizeChar
     ) where
 
 import           GHC.Int
@@ -124,6 +125,16 @@ prevSkip ba offset = loop (offset `offsetMinusE` sz1)
     loop o
         | isContinuation (primIndex8 ba o) = loop (o `offsetMinusE` sz1)
         | otherwise                       = o
+
+sizeChar :: Char -> CountOf Word8
+sizeChar !c
+    | bool# (ltWord# x 0x80##   ) = CountOf 1
+    | bool# (ltWord# x 0x800##  ) = CountOf 2
+    | bool# (ltWord# x 0x10000##) = CountOf 3
+    | otherwise                   = CountOf 4
+  where
+    !(I# xi) = fromEnum c
+    !x       = int2Word# xi
 
 write :: PrimMonad prim => Mutable (PrimState prim) -> Offset8 -> Char -> prim Offset8
 write mba !i !c
--- a/basement/Basement/Block/Base.hs
+++ b/basement/Basement/Block/Base.hs
@@ -8,6 +8,7 @@ module Basement.Block.Base
     , unsafeNew
     , unsafeThaw
     , unsafeFreeze
+    , unsafeShrink
     , unsafeCopyElements
     , unsafeCopyElementsRO
     , unsafeCopyBytes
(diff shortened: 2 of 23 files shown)


basement/Basement/Cast.hs
```diff
@@ -0,0 +1,151 @@
...
+import           Basement.These
+import           Basement.PrimType (PrimType)
+import           Basement.Types.OffsetSize
+import           Basement.Compat.Natural
+import qualified Prelude (fromIntegral)
+
+-- | `Cast` an object of type a to b.
+--
+-- Do not add instance of this class if the source type is not of the same
+-- size of the destination type. Also keep in mind this is casting a value
```
vincenthz: Some of the instances directly disagree with this. `Word128` -> `Word8` is not the same size. I think I agree with the documentation of Cast.

basement/Basement/Cast.hs
```diff
@@ -0,0 +1,151 @@
...
+instance Cast Word64 Word128 where
+    cast w = Word128 0 w
+instance Cast Word64 Word256 where
+    cast w = Word256 0 0 0 w
+
+instance Cast Word Word64 where
+    cast = wordToWord64
+
+
+instance Cast (Block.Block a) (Block.Block Word8) where
```
vincenthz: :+1:

basement/Basement/From.hs
```diff
@@ -191,7 +192,9 @@ instance From (Maybe a) (Either () a) where
 instance From (CountOf ty) Int where
     from (CountOf n) = n
 instance From (CountOf ty) Word where
-    from (CountOf n) = from n
+    -- here it is ok to cast the underlying `Int` held by `CountOf` to a `Word`
```
vincenthz: documentation talk about cast, probably 'convert' would be less confusing considering the Cast class

basement/Basement/Alg/Native/UTF8.hs
```diff
@@ -125,6 +126,16 @@ prevSkip ba offset = loop (offset `offsetMinusE` sz1)
         | isContinuation (primIndex8 ba o) = loop (o `offsetMinusE` sz1)
         | otherwise                       = o
 
+sizeChar :: Char -> CountOf Word8
+sizeChar !c
```
vincenthz: might be a good idea to re-use/re-purpose basement/Basement/UTF8/Helper.hs `charToByte`

basement/Basement/Block/Base.hs
```diff
@@ -321,6 +327,17 @@ unsafeCopyBytesRO (MutableBlock dstMba) (Offset (I# d)) (Block srcBa) (Offset (I
...
+-- | Copy a number of bytes from a Ptr to a MutableBlock with specific byte offsets
+unsafeCopyBytesPtr :: forall prim ty . PrimMonad prim
+                   => MutableBlock ty (PrimState prim) -- ^ destination mutable block
+                   -> Offset Word8                     -- ^ offset at destination
+                   -> Ptr ty                           -- ^ source block
+                   -> CountOf Word8                    -- ^ number of bytes to copy
+                   -> prim ()
+unsafeCopyBytesPtr (MutableBlock dstMba) (Offset (I# d)) (Ptr srcBa) (CountOf (I# n)) =
+    primitive $ \st -> (# copyAddrToByteArray# srcBa dstMba d n st, () #)
+{-# INLINE unsafeCopyBytesPtr #-}
```
vincenthz: could use unsafeCopyBytesPtr in:
```
copyFromPtr ..
copyNative (MutableBlock mba) = primitive $ \st -> (# copyAddrToByteArray# src# mba od# bytes# st, () #)
```

basement/Basement/Block/Builder.hs
```diff
@@ -0,0 +1,138 @@
...
+import           Data.Proxy
+
+newtype Action = Action
+    { runAction_ :: forall prim . PrimMonad prim
+                 => MutableBlock Word8 (PrimState prim)
+                 -> Offset Word8
+                 -> prim (Offset Word8)
+    }
+
+data Builder = Builder Action (CountOf Word8)
```
vincenthz: it doesn't make a difference in runtime, but I'ld probably put the size as a first parameter.

also I would make it strict, I don't see any good use of lazyness here.

basement/Basement/Block/Builder.hs
```diff
@@ -0,0 +1,138 @@
...
+    mempty = empty
+    {-# INLINE mempty #-}
+    mappend = append
+    {-# INLINABLE mappend #-}
+
+-- | create an empty builder
+--
+-- this does nothing, build nothing, take no space (in the resulted block)
+empty :: Builder
+empty = Builder (Action $ \_ _ -> pure 0) 0
```
vincenthz: offset error

basement/Basement/Block/Builder.hs
```diff
@@ -0,0 +1,138 @@
...
+data Builder = Builder Action (CountOf Word8)
+
+instance Semigroup Builder where
+    (<>) = append
+    {-# INLINABLE (<>) #-}
+instance Monoid Builder where
+    mempty = empty
+    {-# INLINE mempty #-}
+    mappend = append
+    {-# INLINABLE mappend #-}
```
vincenthz: I think adding a mconcat is useful here. it's just a review note for an optimisation, no need to solve this straight away

basement/Basement/Block/Builder.hs
```diff
@@ -0,0 +1,138 @@
...
+-- this action is unsafe as there is no guarantee upon the validity of the
+-- content of the built block.
+unsafeRunString :: PrimMonad prim => Builder -> prim String
+unsafeRunString b = do
+    str <- run b
+    pure $ String $ A.UArray 0 (B.length str) (A.UArrayBA str)
+
+-- | add a Block in the builder
+emit :: Block a -> Builder
+emit b = flip Builder size $ Action $ \arr off ->
```
vincenthz: probably at a later point, adding a way to just use a block as is (e.g. for other type of stream builder like list, conduit or file).

The reason is, for a small block, it's fine to inline, but after a certain threshold, you might want to just use the block directly to write to file without copying everything. something to keep in mind

basement/Basement/Block/Builder.hs
```diff
@@ -0,0 +1,138 @@
...
+           -> Offset Word8
+           -> FinalPtr Word8
+           -> prim ()
+    onAddr arr off fptr = withFinalPtr fptr $ \ptr -> B.unsafeCopyBytesPtr arr off ptr size
+
+-- | emit a UTF8 char in the builder
+--
+-- this function may be replaced by `emit :: Encoding -> Char -> Builder`
+emitUTF8Char :: Char -> Builder
+emitUTF8Char c = flip Builder 4 $ Action $ \(MutableBlock arr) off ->
```
vincenthz: 4 ? I think considering what size represent at the moment, must be a bug. it's only when we add max bounds of primitive supports that it will be true.

PR #458, opened 2018-01-11
add weeder test

NicolasDP:


ndmitchell:
You should probably build the tests and benchmarks, then the weeder exceptions will be much lower. If you just pass `--build` to weeder it will do the common settings which typically work.

PR #460, opened 2018-01-12
Fix the build on GHC 7.8

gwils:
This fixes the build on GHC 7.8
The problem is that we're using `pure` given just a `Monad` constraint, which doesn't work before the Applicative-Monad-Proposal went though in GHC 7.10

Currently users on 7.8 can get bad build plans involving foundation 0.0.18. Please issue a hackage metadata revision to tighten foundation 0.0.18's base bound, since it is unbuildable on GHC 7.8. The correct bound for that version would be `base >=4.8 && <5`. This means that users won't get a bad build plan. They won't get a build plan at all, which is a better outcome as it allows the solver to select another version of foundation if it's within their bounds.

Hackage metadata revisions can be issued from here:
https://matrix.hackage.haskell.org/package/foundation#GHC-7.8/foundation-0.0.18

I would appreciate if this fix could be merged and released as part of 0.0.19 to reinstate GHC 7.8 support.

Thanks!

### DIFF
--- a/Foundation/Exception.hs
+++ b/Foundation/Exception.hs
@@ -9,7 +9,7 @@ import Control.Exception (Exception, SomeException)
 import Foundation.Monad.Exception
 
 finally :: MonadBracket m => m a -> m b -> m a
-finally f g = generalBracket (pure ()) (\() a -> g >> pure a) (\() _ -> g) (const f)
+finally f g = generalBracket (return ()) (\() a -> g >> return a) (\() _ -> g) (const f)
 
 try :: (MonadCatch m, Exception e) => m a -> m (Either e a)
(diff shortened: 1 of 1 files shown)


vincenthz:
Thanks @gwils for the PR.

as a general statement, I'm open to keep 7.8 in the working list, but we're really reaching the EOL for 7.8 and very soon for 7.10 toon, so personally I won't put too much efforts in supporting this; If anyone want this to keep happening, then it's probably a good idea to help just like you did :)

w.r.t to `return` vs `pure`, using `pure` is part of an ongoing efforts to drop `return` completely, which I expect will happen to the base's Monad class at some point too. this is effectively making it annoying when the AMP is not there.

There won't be any revisions though

Issue #271, opened 2017-04-28
Improve upper/lower


ndmitchell:
Following on from #261, the current upper/lower functions are character oriented, but real Unicode case conversion is string orientated. The list of cases where that isn't equivalent are at http://unicode.org/Public/UNIDATA/CaseFolding.txt and http://unicode.org/Public/UNIDATA/SpecialCasing.txt. These are dealt with properly by the `text` package.

I suspect this makes little difference to most users, so a low priority one.

DavidM-D:
Text does this by generating a big chunk of code derived from those text files. This seems like the simplest most portable way of doing things, if a little ugly, does anyone have an issue with me doing it that way?

ndmitchell:
@DavidM-D yes, that was what I had in mind.

vincenthz:
@DavidM-D no, that's the right way (albeit ugly) and allow to update easily when unicode revisions are published.

DavidM-D:
Cool, I'll get on that this weekend

ndmitchell:
At the same time as this semantic improvement it would be good to have an algorithm that scanned the string without allocating until it came to a character that needed conversion - then it could just return the input string if there were no characters that needed converting.

PR #461, opened 2018-01-12
Unicode Case Mapping

NinjaTrappeur:
Fixes issue #271.

This PR contains two features:

1. A script that read Unicode case mapping tables and generates a CaseMapping.hs file.
2. The caseConvert function which uses those mapping in order to both upper and lower case a UTF-8 string.

# CaseMapping.hs Generator

This part has been inspired by the text package. Basically, I just translated their parser using the foundation's Parse semantics.

This generator reads both CaseFolding.txt and SpecialCasing.txt files and generate an Haskell equivalent at basement/Basement/String/CaseMapping.hs .

I included a generateCaseMapping.sh convenience script. I had some trouble finding the right stack flags, you can see this file more as a bit of documentation than anything else :)

This script should be manually run for each Unicode revision: usually once a year, around June.

# Upper and Lower Case Conversion

The conversion is done in two pass:

* A first pass (see the caseConvertNBuff function) where we check:
   * the new buffer size needed for the case conversion.
   * if the String needs any modification for this case conversion. (See @ndmitchell [comment](https://github.com/haskell-foundation/foundation/issues/271#issuecomment-340308264))
* A second pass (see the caseConvert function) where we allocate the new buffer (if needed) and fill it using the case conversion function applied on the source string.

# Possible Issues

* The parser will not be built by the CI, hence, it could break at any moment in the future. Have we any way to force the build of this module by Travis? How should we do it?
* I did not benchmarked this change yet.

# Side Notes

This is my first contribution to foundation, I am totally out of my comfort zone here. I think this PR should be **carefully** reviewed before even thinking about merging it, especially the caseConvertNBuff function which could lead to some memory problems if not implemented right.

I did this for training purposes, any kind of feedback is more than welcome :)


ndmitchell:
Awesome! Thanks for your work on this. I haven't yet reviewed but will do so. 

The way I would handle updating and keeping the script alive would be to have the generator run every time in the CI and at the end of the CI do a git diff and fail if there are changes - indicating the generator needs rerunning. The cost to Vincent is that the CI will fail once a year for reasons that are not his fault, but it does serve as a very handy reminder.

basement/Basement/String.hs
```diff
@@ -1317,15 +1318,62 @@ decimalDigitsPtr startAcc ptr !endOfs !startOfs = loop startAcc startOfs
 {-# SPECIALIZE decimalDigitsPtr :: Int -> Addr# -> Offset Word8 -> Offset Word8 -> (# Int, Bool, Offset Word8 #) #-}
 {-# SPECIALIZE decimalDigitsPtr :: Word -> Addr# -> Offset Word8 -> Offset Word8 -> (# Word, Bool, Offset Word8 #) #-}
 
+-- | A unicode string size may increase during a case conversion operation.
+--   This function calculates the new buffer size for a case conversion.
+--   Returns Nothing if no case conversion is needed.
```
ndmitchell: Can the size decrease as well?

basement/Basement/String.hs
```diff
@@ -1317,15 +1318,62 @@ decimalDigitsPtr startAcc ptr !endOfs !startOfs = loop startAcc startOfs
...
+caseConvertNBuff op s@(String ba) = runST $ Vec.unsafeIndexer ba go
+  where
+    !sz = size s
+    !end = azero `offsetPlusE` sz
+    go :: (Offset Word8 -> Word8) -> ST st (Maybe (CountOf Word8))
+    go getIdx = loop (Offset 0) 0 False
+      where
+        !nextI = nextWithIndexer getIdx
+        eSize !e = if e == '\0' 
+                      then 0
```
ndmitchell: Why is 0 of size 0 and not 1?

basement/Basement/UTF8/Types.hs
```diff
@@ -34,6 +36,9 @@ newtype StepDigit = StepDigit Word8
 -- | Step when processing ASCII character
 newtype StepASCII = StepASCII Word8
 
+-- | Specialized tuple used for case mapping.
+data CM = CM {-# UNPACK #-} !Char {-# UNPACK #-} !Char {-# UNPACK #-} !Char deriving (Eq)
```
ndmitchell: Should these be Char or Word8?

tests/Test/Foundation/String.hs
```diff
@@ -59,6 +59,14 @@ testStringCases =
               "this is only a simple string but quite longer than the 64 bytes used in the modified UTF8 parser"
               "this is only a simple string but quite longer than the 64 bytes used in the modified UTF8 parser"
         ]
+    , Group "CaseMapping" 
+         [ CheckPlan "a should capitalize to A" $ validate "a" $ upper "a" == "A"
+         , CheckPlan "b should capitalize to B" $ validate "b" $ upper "b" == "B"
+         , CheckPlan "B should capitalize to B" $ validate "B" $ upper "B" == "B"
+         , CheckPlan "é should capitalize to É" $ validate "é" $ upper "é" == "É"
+         , CheckPlan "ß should capitalize to SS" $ validate "ß" $ upper "ß" == "SS"
+         , CheckPlan "ﬄ should capitalize to FFL" $ validate "ﬄ" $ upper "ﬄ" == "FFL"
```
ndmitchell: Can you please add a random test that capitalising is idempotent or something? Anything that hammers on long strings will show up GC and segfault issues which are the most likely problems.

scripts/caseMapping/CaseMapping.hs
```diff
@@ -0,0 +1,41 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+module CaseMapping where
```
ndmitchell: I always require an explicit export list. Not sure how Vincent feels though (if he does then an HLint rule could enforce it)

basement/Basement/UTF8/Types.hs
ndmitchell: Please document the invariant that it's isomorphic to `[Char]` and that all trailing `Char` must be `\0`.

basement/Basement/String.hs
```diff
@@ -1317,15 +1318,62 @@ decimalDigitsPtr startAcc ptr !endOfs !startOfs = loop startAcc startOfs
...
+                      else charToBytes (fromEnum e)
+        loop !idx ns convert
+            | idx == end = if convert
+                             then return $ Just ns
+                             else return Nothing
+            | otherwise = do
+                let !(c, idx') = nextI idx
+                    !cm@(CM c1 c2 c3) = op c 
+                    !cSize = (eSize c1) + (eSize c2) + (eSize c3)
```
ndmitchell: Unnecessary brackets

basement/Basement/String.hs
```diff
@@ -1317,15 +1318,62 @@ decimalDigitsPtr startAcc ptr !endOfs !startOfs = loop startAcc startOfs
...
+                    !nconvert = convert || not ((c1 == c) && (c2 == '\0') && (c3 == '\0'))
```
ndmitchell: Do you need to test `c3`? If `c2 == 0` doesn't that imply `c3 == 0`?

basement/Basement/String.hs
```diff
@@ -1317,15 +1318,62 @@ decimalDigitsPtr startAcc ptr !endOfs !startOfs = loop startAcc startOfs
...
+    !end = azero `offsetPlusE` sz
+    go :: (Offset Word8 -> Word8) -> ST st (Maybe (CountOf Word8))
+    go getIdx = loop (Offset 0) 0 False
+      where
+        !nextI = nextWithIndexer getIdx
+        eSize !e = if e == '\0' 
+                      then 0
+                      else charToBytes (fromEnum e)
+        loop !idx ns convert
+            | idx == end = if convert
```
ndmitchell: Suggest renaming `convert` to `changed` - you are always doing the conversion, the question is if things change during that process.

basement/Basement/String.hs
```diff
@@ -1317,15 +1318,62 @@ decimalDigitsPtr startAcc ptr !endOfs !startOfs = loop startAcc startOfs
...
+                    !cSize = (eSize c1) + (eSize c2) + (eSize c3)
+                    !nconvert = convert || not ((c1 == c) && (c2 == '\0') && (c3 == '\0'))
+                loop idx' (ns + cSize) nconvert
+
+-- | Convert a 'String' 'Char' by 'Char' using a case mapping function. 
+caseConvert :: (Char -> CM) -> String -> String
+caseConvert op s@(String ba) 
+  = case nBuff of
+      Nothing -> s
+      (Just nLen) -> runST $ unsafeCopyFrom s nLen go
```
ndmitchell: Unnecessary brackets.

basement/Basement/String.hs
```diff
@@ -1317,15 +1318,62 @@ decimalDigitsPtr startAcc ptr !endOfs !startOfs = loop startAcc startOfs
...
+  where
+    !nBuff = caseConvertNBuff op s
+    go :: String -> Offset Char -> Offset8 -> MutableString s -> Offset8 -> ST s (Offset8, Offset8)
+    go src' srcI srcIdx dst dstIdx = do
+      let !(CM c1 c2 c3) = op c 
+      dstIdx'    <- writeChar c1 dstIdx 
+      dstIdx''   <- writeChar c2 dstIdx'
+      nextDstIdx <- writeChar c3 dstIdx''
```
ndmitchell: if c2 is `0` can't you shortcut testing c3? It makes the code less clean, but it's so very common that it's probably worth it.

basement/Basement/String.hs
ndmitchell: Also if c2 == 0 we can avoid adding in c3, which is going to be a saving in the common case.

basement/Basement/UTF8/Types.hs
ndmitchell: Suggest instead using `CM Int Int Int` with `-1` as the sentinel, since the rest of the foundation stuff is Int not Char for the UTFsize etc.

tests/Test/Foundation/String.hs
ndmitchell: Given all the sentinels around 0 please make sure you test both `\0a` and `\0\0` (one that requires transforming and one that does not). I suspect they are broken.

ndmitchell:
I will say this patch looks really really good - my requested changes (which @vincenthz really needs to confirm are the direction he wants to take things) are pretty minor - looks very nice. I definitely owe you a beer.

basement/Basement/String.hs
NinjaTrappeur: Indeed, it could!

scripts/caseMapping/CaseMapping.hs
NinjaTrappeur: Oops, my bad, I made that during the prototyping phase, completely forgot to add it when cleaning...

tests/Test/Foundation/String.hs
NinjaTrappeur: Great idea!

NinjaTrappeur: Alright, I added the property test.

I am a bit confused by your second comment thought. Isn't `\x0a` LF and \0\0 two nulls char? I am assuming you would like a test case where the heading bytes are 0s, would `upper à == À` (`upper \x00e0 == \x00c0`)? 

I am maybe missing the point here...

NinjaTrappeur:
Thanks for this quick review Neil!

basement/Basement/String.hs
NinjaTrappeur: Fixed.

tests/Test/Foundation/String.hs
ndmitchell: I meant `['\0','a']`. Specifically, try a test where nothing in the string converts (so it fails out early), and a test where they do convert (so you test the conversion path too).

basement/Basement/String.hs
NinjaTrappeur: Fixed.

ndmitchell:
That's a question for @vincenthz (although my general feeling is if people want squashed commits they can do it with the merge so no reason to get the contributors to do it)

ndmitchell:
Do let me know when you've done all the changes you are planning on doing and I'll re-review.

basement/Basement/String.hs
NinjaTrappeur: Indeed it does :)

Fixed.

ndmitchell: The whole expression might be clearer as `changed' = changed || c1 /= c || c2 /= 0`

basement/Basement/String.hs
NinjaTrappeur: Fixed.

basement/Basement/String.hs
NinjaTrappeur: It sadly makes the code quite messy. But your're right, the common case should be short.

Fixed, if the code looks too messy, I can revert this change.

tests/Test/Foundation/String.hs
NinjaTrappeur: Right, got it!

Done.

NinjaTrappeur:
I just posted the changes.

The only change I did not push was the CM elements type change to Int. I'll wait for @vincenthz feedback to make my mind on that. It is a minor change anyways, we just need to add a toEnum call on the code generator side and a fromEnum call on the library side. I'll add this change tomorrow night if needed.

Until then, thanks again for the review and enjoy your weekend!

ndmitchell:
Ah, not `"0a"` but `"\0a"` - so a null character followed by `a`. My suspicion is that the code is wrong for strings containing `\0`, and that you'll have to move to Int so you can have access to a sentinel value (e.g. `-1`).

basement/Basement/String.hs
```diff
@@ -1317,15 +1318,68 @@ decimalDigitsPtr startAcc ptr !endOfs !startOfs = loop startAcc startOfs
...
+                      else charToBytes (fromEnum e)
+        loop !idx ns changed 
+            | idx == end = if changed
+                             then return $ Just ns
+                             else return Nothing
+            | otherwise = do
+                let !(c, idx') = nextI idx
+                    !cm@(CM c1 c2 c3) = op c 
+                    !cSize = if c2 == '\0' -- if c2 is empty, c3 will be empty as well.
+                              then eSize c1
```
ndmitchell: For `c1` you can just call `charToBytes` directly and avoid one test against `\0`.

basement/Basement/String.hs
```diff
@@ -1317,15 +1318,68 @@ decimalDigitsPtr startAcc ptr !endOfs !startOfs = loop startAcc startOfs
...
+          then return dstIdx'
+          else do
+            dstIdx''  <- writeChar c2 dstIdx'
+            writeChar c3 dstIdx''
+      return (nextSrcIdx, nextDstIdx)
+          where
+            !(Step c nextSrcIdx) = next src' srcIdx
+            writeChar cc wIdx =
+                if cc == '\0'
+                    then return wIdx
```
ndmitchell: You can avoid the \0 test for c1

basement/Basement/String.hs
```diff
@@ -1317,15 +1318,68 @@ decimalDigitsPtr startAcc ptr !endOfs !startOfs = loop startAcc startOfs
...
+    go :: String -> Offset Char -> Offset8 -> MutableString s -> Offset8 -> ST s (Offset8, Offset8)
+    go src' srcI srcIdx dst dstIdx = do
+      let !(CM c1 c2 c3) = op c 
+      dstIdx' <- write dst dstIdx c1
+      nextDstIdx <- 
+        if c2 == '\0' -- We don't want to check C3 if C2 is empty.
+          then return dstIdx'
+          else do
+            dstIdx''  <- writeChar c2 dstIdx'
+            writeChar c3 dstIdx''
```
ndmitchell: Personally I'd just keep reusing `dstIdx` shadowing it as that eliminates a nasty class of failures where you get the wrong number of primes.

NinjaTrappeur:
Oh yes! I finally get the problem! 

Indeed, it feel like the change to Int is mandatory...

basement/Basement/String.hs
ndmitchell: Maybe `writeMaybeChar`?

NinjaTrappeur:
As you can see, handling \0 is not problematic anymore. This is a side effect of the latest changes (we do not check C1 anymore). 

It still does feel like a time bomb to me, we need to change CM inner type to Int. 

I don't have the time to do it right now, it'll be done until Monday.

scripts/caseMapping/CaseMapping.hs
vincenthz: I'm fine either way, since it's a on the side script

basement/Basement/String.hs
```diff
@@ -1317,15 +1318,68 @@ decimalDigitsPtr startAcc ptr !endOfs !startOfs = loop startAcc startOfs
...
+          else do
+            dstIdx  <- writeValidChar c2 dstIdx
+            writeValidChar c3 dstIdx
+      return (nextSrcIdx, nextDstIdx)
+          where
+            !(Step c nextSrcIdx) = next src' srcIdx
+            writeValidChar cc wIdx =
+                if cc == '\0'
+                    then return wIdx 
+                else do
```
vincenthz: style nitpick: the usual style is:

```haskell
if x
    then a
    else b
```

scripts/caseMapping/CaseMapping.hs
```diff
@@ -0,0 +1,42 @@
...
+main = do
+  psc <- parseSC "SpecialCasing.txt"
+  pcf <- parseCF "CaseFolding.txt"
+  scs <- case psc of
+           Left err -> putStrLn (show err) >> undefined
+           Right sc -> return sc
+  cfs <- case pcf of
+           Left err -> putStrLn (show err) >> undefined
+           Right cf -> return cf
+  h <- openFile ("../../basement/Basement/String/CaseMapping.hs") WriteMode
```
vincenthz: might be easier to have the haskell program write to stdout and redirect the stdout to the file when wanting to rewrite

vincenthz:
nice work @NinjaTrappeur ! also thanks for the review @ndmitchell  !

vincenthz:
I've squashed the commit list since @NinjaTrappeur requested it, but otherwise I don't have a problem to include the history (even if not perfect) in a normal merge. I believe there was no other stuff on @ndmitchell list of things (despite still being in red condition, let use @ndmitchell if I forgot anything). Thanks again @NinjaTrappeur

ndmitchell:
The main thing we need to know is your approach to generation. Mine would be run in ever travis run, fail if there are changes, which means it will fail once a year. Other people have different approaches.

ndmitchell:
And thinking of it, a test showing it is equal to `text` package would give a lot of confidence.

NinjaTrappeur:
I agree with @ndmitchell about the travis part:

* it will run periodically the script, preventing it being outdated.
* it will act as a heads up for every Unicode release. I just realized the text package is 1 release late.

I was actually just writing that part [here](https://github.com/NinjaTrappeur/foundation/commit/e7ae375f3d840907334c32ccf9f91f218eda52c3).

It seems to fail on GHC < 8.2, it needs some adjustments. I'll have a deeper look on that tomorrow.

I'll also have a look on a test showing it is equal to `text`. I think it will also interesting to benchmark the two implementation and see if everything is alright performance wise.

@vincenthz Do you mind if I open a new PR where I can submit those changes?

vincenthz:
@NinjaTrappeur I don't mind at all, the more the merrier.

As to a testing plan, I don't mind too much, the more automated the less likely it's going to create out-of-date problem, but also would like to make sure we still have fast tests (which means minimal extra/optional dependencies).

Showing equivalence to text is useful (provided that it's not out of date) but it's probably a good idea to limit it to the edge travis run which depends on text and bytestring already.

PR #466, opened 2018-01-21
Foreign/Native-PrimArray Algorithms: avoid computing 'endIndex' twice

ciderale:
This is a follow up on PR #434 which got partially merged.

`findIndexElem` (and related functions) have been changed to return `Offset -1` (rather than `Maybe (Offset ty)`) when no matching element has been found. This avoids the use of `Maybe` (as originally done in #434) but still avoids computing the `endIndex` multiple times (c.f. UArray.findIndex). It makes the assumption that `-1` is never a valid index.

Moreover, I adjusted the implementation of `revFindIndexElem` and `revfindIndexPredicate`. The initial bound check (that the array is not empty) is unified with the loop bound check (that the processing reached the start).

### DIFF
--- a/basement/Basement/Types/OffsetSize.hs
+++ b/basement/Basement/Types/OffsetSize.hs
@@ -17,6 +17,7 @@ module Basement.Types.OffsetSize
     ( FileSize(..)
     , Offset(..)
     , Offset8
+    , sentinel
     , offsetOfE
     , offsetPlusE
     , offsetMinusE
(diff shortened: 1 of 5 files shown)


basement/Basement/Types/OffsetSize.hs
```diff
@@ -86,6 +87,8 @@ type Offset8 = Offset Word8
 newtype Offset ty = Offset Int
     deriving (Show,Eq,Ord,Enum,Additive,Typeable,Integral,Prelude.Num)
 
+sentinel = Offset (-1)
```
vincenthz: It would make sense as a pattern synonym instead of just a lvalue

vincenthz:
The sentinel / pattern synonym can be investigated in another PR. the Offset/CountOf API need some further cleanup at this stage

vincenthz:
Thanks, looking good !

PR #468, opened 2018-01-22
Foreign/Native String/UTF8 Code Deduplication

ciderale:
This PR adapts all String/UTF8 functions defined in Basement/Alg/{Foreign,Native} to the mechansim introduced in PR #434. With this change, the Foreign and Native subfolder in Basement/Alg disappear.
Benchmarks look comparable. I further compared the core output to avoid performance penalties. The code should thus behave as before.

Additionally, this PR fixes a minor bug in String.breakEnd which did not account for a possible start offset in the representation. Hence `breakEnd (=='c') $ drop 1 ("abcdef"::String)` returned `(“bcd","ef”)` rather than `(“bc”, “def”)`. 

There are still various functions which have duplicate implementations for Native/Foreign memory (or functions that do not build on the 'onBackend..' approach). I didn't not address them (yet) to keep the changes in this PR as small as possible. This PR already has quite some potential for merge conflicts. After this has been merged we can start converting the remaining duplicates (and the discussion in #466 ), which should be more local changes. @vincenthz I'm looking forward to your feedback.


basement/Basement/UTF8/Base.hs
```diff
@@ -104,12 +104,12 @@ sToList (String arr) = Vec.onBackend onBA onAddr arr
 sToListStream (String arr) k z = Vec.onBackend onBA onAddr arr
   where
     (Vec.ValidRange !start !end) = Vec.offsetsValidRange arr
-    onBA (BLK.Block ba) = loop start
+    onBA ba@(BLK.Block _) = loop start
```
vincenthz: is this `ba@(BLK.Block _)` not equivalent to `!ba` ?

vincenthz:
looking good, both this and #466. is there a specific order to merge those 2 PRs, for some reason it looks like they are addressing similar code path, would you mind merging them both yourself ?

ciderale:
The two commits should not conflict when merging, but better merge this one (#468) first as it is the bigger one that I prefer not having to double check again. Moreover, I will have a look at your suggestion in #466 about the pattern synonym so there might be more commits to come there.

basement/Basement/UTF8/Base.hs
ciderale: You are most likely right.

There were other places where a pattern like this (but probably with a bang `!` on the value within the Block) that made a (significant) difference in the produced core. I will address that when combining the "onBA" and "onAddr" in a next step.

PR #469, opened 2018-01-24
System package

vincenthz:


NicolasDP:
good idea, however the directory starts to be a little messy now. What would you think of moving foundation in a sub folder too ?

NicolasDP:
@vincenthz , would you mind if I were to rebase that branch and move all foundation into a `foundation` sub-directory too ?

Issue #467, opened 2018-01-21
basement-0.0.6  does not compile on OpenIndiana (Open Solaris)


asyropoulos:
I tried to compile basement-0.0.6 on OpenIndiana with ghc-8.2.2 and it failed. I discoved that the following patch solves the problem:

diff -Naur basement-0.0.6/Basement/Terminal/Size.hsc basement-0.0.6.new/Basement/Terminal/Size.hsc
--- basement-0.0.6/Basement/Terminal/Size.hsc   2017-11-11 19:03:59.000000000 +0000
+++ basement-0.0.6.new/Basement/Terminal/Size.hsc       2018-01-21 15:03:03.899338329 +0000
@@ -20,6 +20,9 @@
 #include <windows.h>
 #elif defined FOUNDATION_SYSTEM_UNIX
 #include <sys/ioctl.h>
+#ifdef __sun
+#include <sys/termios.h>
+#endif
 #endif

 #include <stdio.h>

maybe you should include this patch in the source distribution.

Apostolos Syropoulos

vincenthz:
Hi @asyropoulos, we're not supporting openindiana, since no current maintainers uses such a system, but if you put this patch in PR form then it will likely be accepted in the tree.

asyropoulos:
Hello!
Could you please let me know how can I bring the patch in PR form? I am not really sure what PR form is.

NicolasDP:
hi @asyropoulos, thanks for reporting the issue. **PR** here means **P**ull **R**equest. [Github provides documentations here](https://help.github.com/categories/collaborating-with-issues-and-pull-requests/) about pull requests.

NicolasDP:
the patch has been applied, thanks.

PR #472, opened 2018-01-26
add support for OpenIndiana

NicolasDP:
this commit has been initially written by @asyropoulos in #467


NicolasDP:
@asyropoulos , I would love if you could double check this is the change you wish to see upstreamed ?

thanks

PR #473, opened 2018-01-26
move the linking test on a different build target

NicolasDP:
this problem originally only impacted windows build so reduce it
to windows target only.

fix #412


NicolasDP:
relates to #360 (the PR that introduced the test moved in different test script)
relates to [comment (#326)](https://github.com/haskell-foundation/foundation/issues/326#issuecomment-309219955) which is the original source of this test

ndmitchell:
What's the motivation for this? Having it Windows only might cause issues to be missed in future.

NicolasDP:
so here we are solving 2 things:

* Nix users are having issues building foundation as reported by #412 ; So we need to address that.
* The linking problem, is not a _normal_ test as per say. It is testing that the compiler/linker is actually checking linking of the package.

You are right about not performing that test only on windows. I have enabled the test (via a flag) on both linux, osx and windows build (travis and appveyor).

ndmitchell:
Ok, makes a lot of sense - thanks!

PR #479, opened 2018-02-09
BlockN operators and helpers

NicolasDP:
add missing `withPtr` and `withMutablePtr` for `BlockN` and `MutableBlockN`

this will allow us to add support in `memory` for `ByteArrayAccess` on `BlockN`


vincenthz:
looking good :+1:

PR #480, opened 2018-02-12
add Bits and BitOps type class

NicolasDP:


basement/Basement/Bits.hs
```diff
@@ -0,0 +1,486 @@
...
+import qualified Data.Bits as OldBits
+import Data.Maybe (fromMaybe)
+import Data.Proxy
+import GHC.Base hiding ((.))
+import GHC.Prim
+import GHC.Types
+import GHC.Word
+import GHC.Int
+
+-- | type level tag to represet a bit.
```
vincenthz: typo represent

basement/Basement/Bits.hs
```diff
@@ -0,0 +1,486 @@
...
+--
+-- Mainly used for offsets and sizes
+type Bit = Bool
```
vincenthz: This is really consuming some good symbol estate for no good reason, this need to be removed

basement/Basement/Bits.hs
```diff
@@ -0,0 +1,486 @@
...
+
+-- | operation over finit bits
```
vincenthz: typo finit

basement/Basement/Bits.hs
```diff
@@ -0,0 +1,486 @@
...
+class PrimBitOps bits where
```
vincenthz: I don't think it make sense to call it PrimBitOpts, I'ld rather see FiniteBitsOps

basement/Basement/Bits.hs
```diff
@@ -0,0 +1,486 @@
...
+
+-- | operation over bits
+class BitOps bits where
+    (.&.)     :: bits -> bits -> bits
+    (.|.)     :: bits -> bits -> bits
+    (.^.)     :: bits -> bits -> bits
+    (.<<.)    :: bits -> CountOf Bit -> bits
+    (.>>.)    :: bits -> CountOf Bit -> bits
+    -- | reverse all bits in the argument
+    not_      :: bits -> bits
```
vincenthz: `not_` a good name. `bitwiseCompletement` or `bitwiseNegate` make more sense. also this is a finite operation, so it shouldn't be in this class

ndmitchell: For info, I like `invert` as the phrases like bitwise are a bit redundant in the Bits class and negate/complement are too overloaded.

vincenthz: I agree that in the declaration side it's obvious, but on the other hand in the caller, that make it obvious that you're talking about a bitwise operation. I can't really think of any term that is not overloaded here (negate, completement, or invertion)

ndmitchell: It's the length of the identifier I don't like, since both `bitwise` and `Complement` are long. Perhaps `bitflip`?

vincenthz: `bitflip` (or maybe `bitFlip`) sounds fine; I justified the longer name just because unless you're doing bitwise stuff all the time, it's not very common in code.

basement/Basement/Bits.hs
NicolasDP: fair enough

basement/Basement/Bits.hs
NicolasDP: thanks for the feedback. I have made the required changes.

PR #484, opened 2018-02-25
add MonadFix to StateT, ReaderT and ExceptT

### LINKED ISSUE #483: Missing MonadFix instance for StateT
Some libraries rely on `mdo` (such as llvm IRBuilder) and are unusable when using foundation. Could you please add MonadFix instances?

NicolasDP:
fix #483


vincenthz:
This need the `AMPMonad` constraint from Basement.Compat.AMP

PR #487, opened 2018-03-11
Drop GHC 7.8 and 7.10

vincenthz:
drop ghc 7.8 and 7.10 support now that 8.4 is out.


NicolasDP:
Hooray !

ndmitchell:
I always get gently sad when old compiler support is removed. We loved them once, and now cast aside. 

More seriously, doesn't Hackage require an upper bound on base to upload?

vincenthz:
@ndmitchell somewhat sad indeed, 7.8 was a good one, but we need to move on, too much time wasted supporting something a handful of people are using now. w.r.t base: not if you put in a conditional. tested : https://hackage.haskell.org/package/udbus

ndmitchell:
Cool, condition true it is :)

vincenthz:
@ndmitchell actually just was being curious how the conditional block the check, looks like it's the buildable false that prevent the check to trigger.

This works too:
`if impl(ghc < 0)
    buildable: false
build-depends: base`

Issue #485, opened 2018-02-25
Optimise upper & lower casing


vincenthz:
Opening a ticket to track future improvement to unicode casing, and centralizing the improvement ideas & experiment

vincenthz:
Replacing the fall through case of upperMapping to special case the 'a' to 'z' range:

```
-upperMapping c = CM (toUpper c) '\0' '\0'
+upperMapping c | c >= 'a' && c <= 'z' = CM (toEnum (fromEnum c + 0x20)) '\0' '\0'
+               | otherwise            = CM (toUpper c) '\0' '\0'
```

This goes now faster than `Text` in the ASCII benchmark, and a tiny bit slower otherwise:

```
types/String/Upper/ascii/String          time                 8.772 μs  
types/String/Upper/ascii/Text            time                 11.12 μs  
types/String/Upper/mascii/String         time                 16.64 μs  
types/String/Upper/mascii/Text           time                 20.27 μs  
```

vincenthz:
Adding a special case to the converter for the ASCII case which is faster to read and write in UTF8 but also faster to lower/upper case:

```types/String/Upper/special casing/String time                 5.979 μs  
types/String/Upper/special casing/Text   time                 6.232 μs  

types/String/Upper/ascii already upper/String time                 3.461 μs  
types/String/Upper/ascii already upper/Text time                 11.17 μs  

types/String/Upper/ascii/String          time                 3.892 μs  
types/String/Upper/ascii/Text            time                 13.83 μs  

types/String/Upper/mascii/String         time                 7.292 μs  
types/String/Upper/mascii/Text           time                 24.40 μs  

types/String/Upper/uni1/String           time                 9.118 μs  
types/String/Upper/uni1/Text             time                 8.437 μs  

types/String/Upper/uni2/String           time                 5.859 μs  
types/String/Upper/uni2/Text             time                 5.565 μs  
```

vincenthz:
adding @ndmitchell @NinjaTrappeur :)

ndmitchell:
Nice!

PR #488, opened 2018-03-12
Optimise casing for ASCII

vincenthz:
related to #485
This massively speed up the ASCII cases, but the unicode speeding up is not finished. The idea is to pre-encode the unicode cases in a pre-processed format (`CharUtf8`) ready to be blitted to memory with just a simple shift per byte compared to multiple shift and masks per byte.

### DIFF
--- a/tests/Test/Basement/UTF8.hs
+++ b/tests/Test/Basement/UTF8.hs
@@ -1,0 +1,17 @@
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE NoImplicitPrelude   #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeFamilies        #-}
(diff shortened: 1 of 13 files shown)


tests/Test/Basement/UTF8.hs
```diff
@@ -0,0 +1,17 @@
...
+    ( tests )
+    where
+
+import Basement.Types.CharUTF8
+import Foundation
+import Foundation.Check
+import Foundation.String
+
+tests = Group "utf8"
+    [ Property "CharUTF8" $ \c -> decodeCharUTF8 (encodeCharUTF8 c) === c
```
NicolasDP: :fireworks:

NicolasDP:
Looking good.

tests/Test/Basement/UTF8.hs
vincenthz: :trollface:

PR #491, opened 2018-03-14
Remove FP Chat slack link

jship:
Discussed with @NicolasDP this morning.  The FP Chat #haskell-foundation channel was a decent experiment, but that slack is way too hectic and message history is not saved.  This PR removes the link to FP chat.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,6 @@ Documentation:
 
 Discuss:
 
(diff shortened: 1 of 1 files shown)


jship:
image

vincenthz:
sadly agreed, the only time I've seen something, it has been removed by GC before I get to reply

NicolasDP:
seems like gitter is the only viable option at the moment

Issue #497, opened 2018-06-21
`Word256` subtraction uses ones' complement?


aherrmann:
Subtraction on `Word256` seems to be using ones' complement instead of two's complement:
```
>>> 0 - 1 :: Word256
115792089237316195423570985008687907853269984665640564039457584007913129639934
>>> complement 0 :: Word256
115792089237316195423570985008687907853269984665640564039457584007913129639935
```
Note, how the outcome of `0 - 1` is one smaller than `complement 0`.

Is that intentional?

aherrmann:
Resolved by #498

PR #498, opened 2018-06-28
Consistent underflow for Word types

aherrmann:
Addresses #497 .

The current behaviour for underflow e.g. `0 - 1` is inconsistent between `Word32/64` (two's complement) and `Word128/256` (ones' complement). This PR changes that to consistently use two's complement.


vincenthz:
Thanks ! good catch

PR #499, opened 2018-07-05
basement: build fix for ghc-8.6

tmcdonell:


vincenthz:
Thanks for sending this as a PR ! But I have to say this fix/workaround makes me sad for haskell

PR #502, opened 2018-08-10
Use HTTPS URL instead of HTTP

nwtgck:
HI, thank you for your great project!

I just change HTTP links to HTTPS for the `foundation` documents. I believe HTTPS is better even if the site is static.


vincenthz:
Thanks !

Issue #518, opened 2019-04-08
Inaccurate package description


andrewthad:
The package description opens with

> A custom prelude with no dependencies apart from base.

However, `foundation` has several non-base dependencies. This should probably be rephrased.

PR #519, opened 2019-04-08
Explicitly bring fail into scope in module with incomplete patterns in do blocks

andrewthad:
Resolves https://github.com/haskell-foundation/foundation/issues/518


vincenthz:
it doesn't fix 518, nor it's valid fix for `fail`

PR #522, opened 2019-06-07
IndexedCollection: fixed findIndex for array types

acowley:


vincenthz:
good catch. looks like it has never been used yet

PR #526, opened 2019-07-14
Make IPv4 parser safe

akshaymankar:
The parser returned `error` if it was passed anything that didn't look like an IP address instead of `ParseError`, thanks to `read` being unsafe. This PR fixes it.


akshaymankar:
Bump.

vincenthz:
Thanks. I tweaked the patch to use TypeApplication and pattern guard but otherwise :+1:

PR #535, opened 2020-03-18
Switch from importing GHC.Prim to GHC.Exts

andrewthad:
In GHC 8.12, unsafeCoerce# is no longer exported by GHC.Prim. Additionally, the documentation for GHC.Prim tells users to import GHC.Exts instead.


vincenthz:
superseded by another patch

PR #538, opened 2020-05-29
FinalPtr: Import GHC.ForeignPtr as qualified

bgamari:
This avoids shadowing of GHC.ForeignPtr.FinalPtr, which will be exposed in GHC 8.12.

PR #537, opened 2020-03-26
Import unsafeCoerce# from GHC.Exts for compatibility with GHC 8.12

sgraf812:
Fixes #536.

PR #540, opened 2020-08-18
Eta convert a lambda to fix GHC 9 type error

utdemir:
Previously this code did not compile on (soon to be released GHC 9). I think it is caused by the [simplified subsumption work](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.0#simplified-subsumption).

I was able to compile `foundation` on GHC 9 with this change alongside with https://github.com/haskell-foundation/foundation/pull/538 and https://github.com/haskell-foundation/foundation/pull/537.


utdemir:
(CI seem to have failed for a different reason)

fromjavatohaskell:
I was able to compile foundation with current master and change from this pull request under ghc-9.0.1

vincenthz:
oh, I missed this PR. I applied effectively the exact same patch from the nix closure description in another ticket

PR #550, opened 2021-03-09
Allow building with ghcjs

hamishmack:


angerman:
Can we please get this merged, and the outright prohibition of ghcjs reverted? How are we supposed to make any progress on ghcjs compatibility wrt to foundation if it refuses to even be included in any form of build plan computation?

@vincenthz pretty please?

vincenthz:
archiving repository

PR #558, opened 2022-01-14
Solve Issue #557, compatibility with Win32 >= 2.12.0

Release-Candidate:
Use `queryPerformanceCounter` instead of `c_QueryPerformanceCounter`, which isn't exported any more by versions of Win32 >= 2.12.0.

I stored the ticks in `startPrecise` in the upper 64 bits of `blk` like before, which isn't necessary any more - just in case someone uses this.

See https://github.com/haskell-foundation/foundation/issues/557


vincenthz:
the CI is in a state of disarray, so merging this on blind faith for now, but thanks !

Issue #569, opened 2022-08-29
rotateL for Word256 seems like it's broken for some cases.


AndreasPK:
See https://gitlab.haskell.org/ghc/ghc/-/issues/22075#note_449902 some of the guards seem to have a infinite loop in their RHS.

Maybe we missed something and it's fine but looked like a bug to me and matt so reporting it here.

PR #570, opened 2022-09-29
Fix infinite loop in rotateL @Word256

ghost:
See https://github.com/haskell-crypto/cryptonite/issues/370, https://gitlab.haskell.org/ghc/ghc/-/issues/22075#note_449902 and https://github.com/haskell-foundation/foundation/issues/569

I haven't tested this, just made the edits on GitHub - Hopefully I got it right!

(If you've got comments, please also ping @axman6)


vincenthz:
archiving repository

